Issue #24, opened 2015-02-28
Incorrect delta output for all paths more than 2 levels deep


tkurki:
See MWV and DBT tests at https://github.com/SignalK/signalk-parser-nmea0183/tree/delta-for-deep-paths/test

tkurki:
Path is only two levels deep and the third level is under value:
```
{
  "context": "vessels.85F33882",
  "updates": [
    {
      "source": {
        "timestamp": "2015-02-28T09:29:44.820Z"
      },
      "values": [
        {
          "path": "environment.depth",
          "value": {
            "belowTransducer": {
              "source": {
                "type": "NMEA0183",
                "src": "DBT",
                "label": "signalk-parser-nmea0183"
              },
              "timestamp": "2015-03-28T09:29:44.000Z",
              "value": 10.83
            }
          }
        }
      ]
    }
  ]
}```

fabdrol:
The issue is coming from multiplexer. I added the tests for the issue to master, along with some other changes and removed the separate branch. I'm in the process of updating multiplexer to fix this issue. Closing for now, if multiplexer doesn't fix it we can always re-open.

tkurki:
As fixing this seems actually pretty simple and doesn't need any Multiplexer changes/rewrite I am reopening.

PR #26, opened 2015-03-22
Remove old failing redundant tests

timmathews:
I don't think we need these anymore, since there are tests for DBT and MWV in parsing.js


tkurki:
Sheesh, things change so fast that tests created a day ago are 'old'..

Seriously: @fabdrol added the tests yesterday 

https://github.com/SignalK/signalk-parser-nmea0183/commit/6b725bca9deb932d7b755ba4e620ecd9b331a56d

to address #24 - which he for some reason closed.

I would have preferred keeping the issue open, adding the tests in a branch and merging & closing when the issue is fixed.

Furthermore instead of a single test I would prefer having tests per sentence, not just a single parsing.js.

timmathews:
So, in reality those are new tests (which fail) and the old tests (in parsing.js) succeed?

tkurki:
Yep, but nevertheless your commit https://github.com/SignalK/signalk-parser-nmea0183/commit/c13196efa5c5c2106a0f59436947799e043175ae was spot on.

PR #35, opened 2015-09-13
New "lite" version

fabdrol:
**NOT READY FOR MERGING (YET)**

This new "lite" version is intended to fix a few fundamental problems with the old parser:
- In-memory SK tree (state), which just increases memory usage without providing any real benefit
- Hard to maintain to keep up with iterations in SK spec: decoding was hardcoded in "codecs"
- Delta updates are second-class citizens

The new "lite" parser does things differently:
- No more state; parsing is done on the fly and nothing is kept in-memory
- Delta messages are first-class citizens and the only message format
- Definitions are kept in a "mappings" file, which makes it easier to add sentences and update the parser
- Improved memory footprint (at least, theoretically. I will benchmark this once it is finished) 

I set up this PR to get feedback whilst I finalise development of the lite version. Three main tasks left:
- Write mappings for other sentences
- Implement AIS decoding (as a separate module?) and decide how it fits in mappings.
- Write new tests for new parser.


mappings.xml
```diff
@@ -0,0 +1,131 @@
...
+      </fieldset>
+      <filter type="if">
+        <field>1</field>
+        <value transform="uppercase">A</value>
+      </filter>
+    </sentence>
+  </mapping>
+
+  <mapping>
+    <path>~/navigation/position/longitude</path>
```
tkurki: Since position.latitude, position.longitude and position.timestamp are not related how do you propose to create just one delta message out of these, or does the code already handle this?



This demonstrates the fact that I find the association order of this mapping a bit cumbersome - the mapping would be much more natural if it were from input (sentence) to output (delta messages eg. Signal K paths).

See also https://github.com/SignalK/n2k-signalk/blob/master/n2kMappings.js (hmm `node` should be renamed to `path`).

fabdrol: I tend to agree, I would have preferred a JSON mapping that is less cumbersome. This is based on Tim's mapping, and the general idea is quite smart: each `<mapping>` has one path, and one or more sentences and PGNs.. and in the future perhaps other mappings.

tkurki: The `value` of the `navigation.position` should be an object like so
```
{
    "context": "vessels.123456789",
    "updates": [{
        "source": {
            "label": "foobar",
            "type": "NMEA2000",
            "pgn": "129025",
            "src": "160"
        },
        "timestamp": "2014-08-15-16:00:21.860",
        "values": [{
            "path": "navigation.position",
            "value": {
                "longitude": 24.9027157,
                "latitude": 60.0387795
            }
        }]
    }]
}
```

I think that the *value* in the delta message should suffice. The delta should read for a human _here are some updates about this context, here's the source & timestamp and here are the values". Primitive values are primitives, while composite values like position and orientation should be objects.

```
{
    "context": "vessels.123456789",
    "updates": [{
        "source": {
            "label": "foobar",
            "type": "NMEA2000",
            "pgn": "129026",
            "src": "160"
        },
        "timestamp": "2014-08-15-16:00:10.999",
        "values": [{
            "path": "navigation.speedOverGround",
            "value": 3.87
        }, {
            "path": "navigation.courseOverGroundTrue",
            "value": 171.8
        }]
    }]
}
```

For primitive values Multiplexer adds the `.value` key: https://github.com/SignalK/signalk-multiplexer-node/blob/master/lib/delta.js#L31

fabdrol: Multiplexer adds it because it wasn't in the pad; is that something we like or do we want to change it? Imho it adds complexity. 
I'll update the parser to compile the value object properly, as you posted!

tkurki: I find delta as posted by me above far more understandable than having `.value` in some paths and not in others. 

If we want everything uniform then all the leaf values should be under `value` key in the schema, so the lat long object would go under `navigation.position.value` and not directly under `navigation.position`. But that is another issue.

fabdrol: I guess there are arguments for both. For now I'll keep things as they are, and won't stir up another discussion ;-)

tkurki: Above you mentioned documentation. Where and how should this be documented? 

I would also love to have a tool that would verify this behavior. Signalk/testclient could easilyl be extended to convert all deltas to full and run verify against that.

fabdrol: I was referring to the "description" parts of the delta.json schema.. Main issue was confusion, as I haven't used delta in a while!

lib/Parser.js
```diff
@@ -0,0 +1,337 @@
...
+
+    return value;
+  };
+
+  Parser.prototype.getValue = function(config, values) {
+    if(_.isObject(config.fieldset)) {
+      return this.getValueFromFieldset(config.fieldset, values);
+    }
+
+    let value = null;
```
tkurki: Are we moving to ES6? Travis doesn't like this out of the box.

fabdrol: I built it using the latest stable version of node (4.0.0) which is a merge between Node.js and io.js. I like the new ES6 features, but I'm not sure it's the best time (I'm having lot's of issues with binary dependencies that don't like 4.0.0)

fabdrol:
@tkurki latest commit updates the incorrect delta format, RMC now comes out like below. Feedback?

```javascript
{
  "context": "vessels.self",
  "updates": [
    {
      "source": {
        "type": "NMEA0183",
        "label": "signalk-parser-nmea0183",
        "sentence": "RMC"
      },
      "values": [
        {
          "path": "navigation.position",
          "value": {
            "latitude": 52.552146666666665,
            "longitude": 5.45415
          }
        },
        {
          "path": "navigation.courseOverGroundTrue",
          "value": 219.63
        },
        {
          "path": "navigation.speedOverGround",
          "value": 2.6082339940859454
        }
      ],
      "timestamp": "2015-05-10T15:08:00.000Z"
    }
  ]
}
```

fabdrol:
I've updated the mapping to be a little bit more concise: 
```xml
<mapping>
    <path>~/navigation/position/longitude</path>
    <sentence id="RMC" filter="uppercase:1 == A">
      <fieldset type="coordinate">
        <field>4</field>
        <field>5</field>
      </fieldset>
    </sentence>
  </mapping>
```
The sentence metadata (it's ID and a filter) are now attributes. The filter is build up like this:
```
uppercase             :   1      ==        A
<optional transformer>:<field><operator><match>
```

fabdrol:
Unless I or someone else comes across a bug, there's only one more thing to do implementation-wise: test with multiple sentences (this requires a change to the "lookup" file). After that, it's all about writing mappings.

fabdrol:
I have updated the lite parser to support more than one sentence per mapping. Three tasks remain:
- Implement functionality to parse AIS sentences
- Implement mappings for all known sentences
- Implement tests for each sentence

Major task, so all help is welcome! I might write a small web app that makes it easier to create mappings, allowing community members to help out.

tkurki:
Source output is not valid according to schema, there is no `sentence` schema.
```
      "source": {
        "type": "NMEA0183",
        "label": "signalk-parser-nmea0183",
        "sentence": "GLL"
      },
```
https://github.com/SignalK/specification/blob/master/schemas/definitions.json#L22

I think we should add `sentence` and `talker` to the delta schema.

tkurki:
Furthermore: when/if we construct full tree and need to handle the classic *multihull case* eg. two depth sounders we need to have enough information in the delta to be able to distinguish between the sources. 

That's where we need both `talker` id (several echosounders multiplexed to same nmea0183 bus with unique talker ids) and `label` (several nmea0183 "buses" with echosounders, same talkerid, distinguishable different `label` values).

So `label` must be meaningful as well, it is not a parser identifier but together with `talker` (and possibly `sentence` if we want to make that distinction) identifies the source of the data.

fabdrol:
@tkurki thanks, there's an issue however. The parser doesn't know anything about the source of the data - it's just piped in between - unless when it can be derived from the data (which is not the case with standard 0183, as far as I know). An option would be to pass the value used in `label` via configuration. 

Re `sentence`: I mainly added that for debugging purposes, so I could see which sentence produced which delta message. I don't think it's meaningful to add it to the delta spec, SK devices don't need to know about what sentence a value came from. I think it would be better to change the name of the key to `_sentence`, indicating that it is "private" and can be safely ignored by servers and consumers alike.

tkurki:
I think that we should include the sentence, if for nothing else then for debugging further down the pipeline. I find it presumptuous to think that nobody will ever need that information.

Correct `label` either has to come from configuration/constructor call or the server needs to fill it in. Either way outputting just the parser is misleading.

How should we proceed?

fabdrol:
Let's do a PR for the sentence, I'll add changes to the code allowing it to accept a `label` in the constructor. 

Regarding todo, essentially there are three things that need doing: (1) AIS support, (2) write mappings, (3) write a test for each mapping. 1 is straightforward, just takes time so no priority. 2 is a shitload of work. We could start writing mappings already, or wait for the tool. 3 should be done in tandem with 2.

tkurki:
I already added using provider id as the label in the server. This way it works the same for all providers and it comes straight from the configuration file.

https://github.com/SignalK/signalk-server-node/commit/d0d855b33831c1adffb07544d5df755e880520dc

tkurki:
@fabdrol what is the status of this PR?

tkurki:
Closing due to inactivity.

joabakk:
Label can be taken from the tag block if available. If you have two sensors connected to kplex with their own interface, you will be able to distinguish.

Issue #38, opened 2016-01-03
SI units


tkurki:
Signal K specification states SI units and nmea0183 parser must be updated to produce SI units.

philgutowski:
@tkurki mind giving a bit more detail about what needs to be changed?

tkurki:
Sure. We recently made the move to use SI units in Signal K. In practice this means that some of the quantities need to be changed so that they are using the correct units. 

A practical example is that angles and directions are currently in degrees in signalk-parser-nmea0183. They need to be converted to radians.

All in all the parser's conversions from nmea0183 need to be checked so that they are in correct units.

See https://github.com/SignalK/specification/blob/master/keyswithmetadata.json

tkurki:
Ping @fabdrol

philgutowski:
Here's a start... https://github.com/SignalK/signalk-parser-nmea0183/pull/39

tkurki:
Closed by #41

Issue #1, opened 2014-04-15
`_transform` doesn't work when a NMEA sentence comes in across multiple lines


fabdrol:
The `_transform` method in `lib/index.js` doesn't work when NMEA sentences come in across multiple lines. Example: I have a GNS 1000 bluetooth GPS receiver at home, which emits NMEA sentences. However, sometimes a sentence comes in across multiple lines. The Transform stream should recognise this by detecting the beginning of a sentence (`$`) and the end `*xx` of a sentence.

I'll take a look at this tonight.

PR #39, opened 2016-01-08
WIP: [Issue #38] Use SI Units

philgutowski:
Will these changes be overwritten when https://github.com/SignalK/signalk-parser-nmea0183/pull/35 gets merged?

### DIFF
--- a/codecs/MWV.js
+++ b/codecs/MWV.js
@@ -45,12 +45,16 @@ Field Number:
 
 var Codec = require('../lib/NMEA0183');
 
+Math.radians = function(degrees) {
+  return degrees * Math.PI / 180;
+};
+
 function convertToWindAngle(self, angle) {
         var numAngle = self.float(angle) % 360;
         if (numAngle > 180 && numAngle <= 360) {
-                return numAngle - 360;
+                return Math.radians(numAngle - 360);
         }
-        return numAngle;
+        return Math.radians(numAngle);
 }
 
 module.exports = new Codec('MWV', function(multiplexer, input) {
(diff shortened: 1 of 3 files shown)


test/parsing.js
```diff
@@ -97,7 +97,7 @@ describe('MWV parser; sentence with true wind data', function() {
                             }
                           },
                           "angleTrueWater": {
-                            "value": -42,
+                            "value": -0.7330382858376184,
```
philgutowski: @tkurki was I right to update these manually?

codecs/MWV.js
```diff
@@ -45,12 +45,16 @@ Field Number:
 
 var Codec = require('../lib/NMEA0183');
 
+Math.radians = function(degrees) {
+  return degrees * Math.PI / 180;
+};
```
philgutowski: Pull this out for use in the other codecs?

tkurki: As there is no real need for monkey patching let's not do that.

Now that I look at this there's https://github.com/SignalK/signalk-parser-nmea0183/blob/master/lib/NMEA0183.js#L66

I'm not terribly fond of the big (and growing) if block, but to stay in line with the rest of the module I'd say conversions should go here.

test/parsing.js
tkurki: Yes. Check also https://github.com/SignalK/signalk-parser-nmea0183/issues/36 - getting the conversions updated to the proper SI units is priority #1, but fixing precision at the same time would not hurt.

codecs/MWV.js
fabdrol: That's where they should go, allowing them to be used across all codecs. @philgutowski do you want to update the rest or shall I go ahead and push a fix for all codecs moving to SI units everywhere?

philgutowski: @fabdrol thanks.  If you have the time, go ahead with it.  This was a good exercise, and I'm starting to better know my way around the system.

PR #44, opened 2016-02-02
Tried to add ROT and RPM

joabakk:
I have added ROT and RPM to my test fork, added these to codecs/index.js and used DBT as example. It seems the parser.on('delta', function(delta) { -function is not working properly and I get timeouts in mocha. Have I forgotten something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,3 +40,9 @@ TODO (SENTENCES)
 - HDG
 - MWD
 - XDR
+
+- ROT
+- RPM
+- VDR
+- VPW
+- VWR
--- a/codecs/ROT.js
+++ b/codecs/ROT.js
@@ -1,0 +1,59 @@
+/* 
+* RPM codec 
+* 
+* @repository https://github.com/signalk/nmea-signalk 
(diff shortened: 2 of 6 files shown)


tkurki:
Your checksums are incorrect and the parser silently ignores the input. Please fix them and everything should start working.

@fabdrol I don't think it should ignore input without any error message.

tkurki:
You can commit the fix to your repo and I can merge this manually - I changed README on January 27th apparently after your fork and replaced the sentence list with a link to the list in code.

joabakk:
Ok, that was embarrassing. Fixed in my repo now.

tkurki:
Merged with #45. I squashed your commits, so nothing left for posterity to wonder about.

Schema has engine instance in the path, so I decided to just hard code one here. That might be a configurable option for Parser in the future (ping @fabdrol ).

For next time: I'd appreciate meaningful commit messages. I guess the ones you used were generated by a tool? Github and git tells what file changes were done, commit message is for a human to understand what actually changed and why. See http://chris.beams.io/posts/git-commit/.

PR #46, opened 2016-02-02
Added support for VDR (current set and drift)

joabakk:
new file:   codecs/VDR.js
	modified:   codecs/index.js
	new file:   test/VDR.js

### DIFF
--- a/test/DBT.js
+++ b/test/DBT.js
@@ -1,15 +1,19 @@
 var chai = require("chai");
 chai.Should();
 chai.use(require('chai-things'));
+var signalkSchema = require('signalk-schema');
 
 
 var nmeaLine = "$IIDBT,035.53,f,010.83,M,005.85,F*23";
 
 describe('DBT', function() {
   it('converts ok', function(done) {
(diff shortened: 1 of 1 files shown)


tkurki:
Current is a composite value in the schema, so the delta should look like this

```
{
  "updates": [{
    "source": {...},
    "timestamp": "2014-08-15-19:00:07.562",
    "values": [{
      "path": "environment.current",
      "value": {
        "setTrue": 2.5793,
        "drift": 0.43
      }
    }]
  }],
  "context": "vessels.urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d"
}
```

tkurki:
I just added validation against the full schema to DBT test with 06e5422331bc8b25326c56010feed2a0e175ced1. Adding that to this test should catch the issue I mentioned above.

In essense this validates the paths and values and not just the json structure of the delta message.

joabakk:
Had to reset my repo. Opening a new PR

PR #51, opened 2016-02-11
Added provision for proprietary sentences

joabakk:
The proprietary sentences will keep the 4 character talker id as "instrument", including "P". This can later be used to distinguish between different protocols with the same "type"


joabakk:
trying again. Sorry, learning the ropes here...

PR #55, opened 2016-03-23
Added support for (deprecated) VWR sentence

joabakk:
Relative wind angle and speed

### DIFF
--- a/codecs/VWR.js
+++ b/codecs/VWR.js
@@ -1,5 +1,5 @@
 /*
-* RPM codec
+* VWR codec
 *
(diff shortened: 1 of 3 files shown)


codecs/VWR.js
```diff
@@ -0,0 +1,74 @@
+/*
+* RPM codec
```
tkurki: These comments don't really serve a purpose I think and can be omitted, as they are frequently forgotten.

joabakk:
Updated the comment. However I agree they could be removed from all sentences, and replaced by a reference to the project and the location, if anything

PR #59, opened 2016-05-20
Multiline ais

tkurki:
Add support for multiline AIS messages.

Improves AIS message handling a bit, so that it checks for existence of individual fields and creates delta accordingly.

The session handling is naive and **will not work for multiple ais streams** because of a shared singleton. The parser needs more changes to support a per parser instance session context.


tkurki:
Fixes #59 

Ping @sailoog and @fulup-bzh

PR #60, opened 2016-08-11
Changed ROT unit from deg/min to rad/s

joabakk:
The original code used deg/min, but the spec says rad/s


joabakk:
will sort out the mess first

PR #61, opened 2016-08-11
Changed ROT from deg/min to rad/s

joabakk:
Spec says rad/s, NMEA outputs deg/min


tkurki:
I would really really appreciate including tests in additions. This time I added one myself: 9279b514062da33231314ccfb9c18477f59989b2

joabakk:
Thank you. The test as was worked, but I agree it was not sufficient.

PR #67, opened 2016-12-02
Support for Seatalk through $STALK sentence

joabakk:
Seatalk1 units should be able to speak NMEA, or at least there are translators available from seatalk1 to NMEA0183, but for more fine grained information, there are some pieces of the puzzle that are not available in NMEA0183. Some units like this one http://www.gadgetpool.de/bestellen/catalog/product_info.php/products_id/54
provide seatalk1 datagrams enclosed in a $STALK sentence. This is an attempt of getting seatalk info from $STALK into SignalK. Now the procedure could be different, and I hope you have some specific input (architecture etc). The STALK.js file may become cumbersomely large if this expands, so it may be necessary to split the different datagram decoders into separate files, but I have not managed to do so. 

The only datagram that has been populated is 0x84, but this provides compass heading, target heading, autopilot mode and rudder angle from seatalk1 autopilots.

### DIFF
--- a/codecs/STALK.js
+++ b/codecs/STALK.js
@@ -37,7 +37,6 @@ where:
 2 			hex   		Last datagram content 
 3 			hex      	Checksum 
 
-$STALK,9C,E1,15,00*4B
 */
 
 var Codec = require('../lib/NMEA0183');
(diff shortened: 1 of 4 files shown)


joabakk:
I'm OK with this now. I have decoded 0x84 and 0x9C datagrams. The other datagrams have direct translation to NMEA0183 and can be added by someone at a later stage. These two are the only datagrams I receive from my ST2000+ autopilot.

PR #68, opened 2016-12-08
Split seatalk datagrams

joabakk:
The seatalk1 datagrams are split out for ease of maintenance/reading and allows direct seatalk datagrams to be decoded without the $STALK wrapper

### DIFF
--- a/test/STALK.js
+++ b/test/STALK.js
@@ -24,7 +24,7 @@ var standby = "$STALK,84,E6,15,00,00,00,00,00,08*1E"
 var auto = "$STALK,84,56,5E,79,02,00,00,00,08*16"
 var wind = "$STALK,84,06,00,00,04,00,00,00,00*63"
 var route = "$STALK,84,06,00,00,08,00,00,00,00*6F"
-var rudder = "$STALK,84,06,00,00,08,00,FE,00,00*5C"
+var rudder = "$STALK,84,06,00,00,08,00,FE,00,00*6C"
 var heading_nineC = "$STALK,9C,51,1E,00*4B"
 
 parser = new(require('../lib/').Parser)();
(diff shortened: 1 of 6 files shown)


tkurki:
Could you please add the Apache copyright notice to the files (and replace my name there with yours?).

joabakk:
Simple enough. Done

tkurki:
The tests fail on my machine  because of missing `sentence` in source:
```
  1) STALK 0x84 heading converted:
     AssertionError: Invalid type: undefined (expected string):/updates/0/source/sentence (0 other errors not reported here)
```

I checked out your branch, rebased it on top of signalk/master, resolved the conflicts by using your versions of codecs/STALK.js and test/STALK.js. You need commit https://github.com/SignalK/signalk-parser-nmea0183/commit/6a8a38e71be7db90b8a3553bc35de13f34f4b0b2 to fix the tests in general.

joabakk:
I might need some help on this. I see the error with mocha, but it tests fine with npm test. All I've done to tune which of the sub-tests work is to move the done() around.
Also, travis ok-ed my branch (got errors, but unrelated) https://travis-ci.org/joabakk/signalk-parser-nmea0183/jobs/182312347

tkurki:
Well, Travis errors this PR, see https://travis-ci.org/SignalK/signalk-parser-nmea0183/builds/182896352

Errors like I got, for missing sentence in source in delta:
```
  1) STALK 0x84 heading converted:
     AssertionError: Invalid type: undefined (expected string):/updates/0/source/sentence (0 other errors not reported here)
```

joabakk:
Ok I'll talk to Travis

joabakk:
Phew, it passes now.

test/STALK.js
```diff
@@ -24,9 +40,9 @@ describe('STALK', function(done) {
         'value': 5.305800926062761
       });
       delta.should.be.validSignalKDelta;
-      done();
```
tkurki: I don't think this can be correct. If the done() call is not triggered on output your test will pass always, even if the parser never emits anything.

joabakk: I had done(); in both the `delta.should..` and `parser.write(...` blocks, but removed the ones in `delta.should....` This is in line with all the other test files.

joabakk:
You are right. I got ut working by inadvertently disabling the test. It did not decode. It seems that this: https://github.com/SignalK/signalk-parser-nmea0183/blob/master/lib/NMEA0183.js#L65 causes problems when another layer is put on the structure.

joabakk:
Not sure I can manage the last hurdle here. Putting this to rest for a while...

PR #72, opened 2016-12-16
Added RPM, ROT and VPW hooks

joabakk:


test/ROT.js
```diff
@@ -0,0 +1,55 @@
...
+
+  it('Converts OK using stream parser', done => {
+    const parser = new Parser
+    const stream = parser.stream()
+
+    stream.on('data', result => {
+      result.should.be.an.object
+      result.should.have.property('delta')
+      result.delta.updates[0].values.should.contain.an.item.with.property('path', 'navigation.rateOfTurn')
+      result.delta.updates[0].values.should.contain.an.item.with.property('value', (35.6 / 180 * Math.PI / 60))
```
fabdrol: You could probably make this test pass by changing this to:

```javascript
// I didn't test this, so that precision of 0.05 might be off. 
result.delta.updates[0].values[0].value.should.be.closeTo((35.6 / 180 * Math.PI / 60), 0.05)
```

test/ROT.js
```diff
@@ -0,0 +1,55 @@
...
+chai.use(require('chai-things'))
+
+describe('ROT', () => {
+
+  it('Converts OK using individual parser', done => {
+    const parser = new Parser
+
+    parser.on('signalk:delta', delta => {
+      delta.updates[0].values.should.contain.an.item.with.property('path', 'navigation.rateOfTurn')
+      delta.updates[0].values.should.contain.an.item.with.property('value', (35.6 / 180 * Math.PI / 60))
```
fabdrol: You could probably make this test pass by changing this to:

```javascript
// I didn't test this, so that precision of 0.05 might be off. 
result.delta.updates[0].values[0].value.should.be.closeTo((35.6 / 180 * Math.PI / 60), 0.05)
```

hooks/RPM.js
```diff
@@ -0,0 +1,54 @@
...
+  try {
+    const delta = {
+      context: 'vessels.self',
+      updates: [
+        {
+          source: tags.source,
+          timestamp: tags.timestamp,
+          values: [
+            {
+              path: 'propulsion.engine1.revolutions',
```
fabdrol: Not sure we can just use `engine1`.. it's an issue, as parser would have to know what engine the value would be coming from. One possible solution would be to make the context configurable - both at a global level (`new Parser({ context: 'vessels.foobar' })`) and a individual sentence level (`parser.parse(sentence, 'vessels.self.propulsion.engine1')`). In the latter case, the hook would need to be smart enough to recognise the context change the delta. 

I'll create a separate issue about this, and we can solve this later.

joabakk:
Will clean up and issue new PR

PR #77, opened 2017-02-01
2.0 version of the nmea0183 parser

fabdrol:
@joabakk and I have completely rewritten the nmea0183 parser. The new parser is more robust (better error handling ánd better error recovery) more and faster. Moreover, the 2.0 version is SK 1.0-ready and fully backwards-compatible. We'd like to merge this into master and publish to NPM. In terms of sentence support: apart from the (incomplete) support for DSC sentence, this version includes everything the old parser had.


fabdrol:
@timmathews @tkurki @canboat @rob42 thoughts, objections?

tkurki:
Have you tested it with the server? Run it for extensive periods?

Since this is a rewrite do you have any relative performance numbers?

Did you update the deps? This would be a great opportunity to do that and start using Greenkeeper.

package.json
```diff
@@ -1,43 +1,44 @@
 {
-  "name": "nmea0183-signalk",
-  "version": "0.2.4",
+  "name": "signalk-parser-nmea0183",
+  "version": "2.0.6",
```
tkurki: Why 2.0.6?

fabdrol:
@tkurki 
1. Yes
2. No (@joabakk how about you?)
3. I do not, but since it doesn't keep a full tree in memory anymore it should at the very least be much lighter on memory (in fact, it doesn't keep anything in memory anymore. It just translates to deltas). I don't have time to do proper benchmarks right now, but I'd welcome the intitiative.
4. Yes, and good idea. I'll look into that. 

More important than the above points however, is that this new version is much easier to maintain and much more robust when it comes to exceptions. Because parsing now happens in a promise chain (and the stream interface is no more than a thin wrapper around that) errors are caught and handled. On top of that, writing hooks for new sentences is much easier since one doesn't need to bother with the multiplexer anymore.

package.json
fabdrol: I started counting from 2 and up, and this is the 6th minor release (in terms of features) of the 2.0.x branch.

joabakk:
Testing now with nmea0183 from log, both with and without tags.

fabdrol:
Whoops, we were too fast; running it longer than a bit makes it fail despite no errors and tests succeeding. Will bug hunt and update.

tkurki:
Have you removed all schema validation tests on purpose? I think they should be part of every test.

tkurki:
The fix & additional test for concurrent AIS parsers needs to included as well: b71b9480ad0b925aec4ccf59e53c779dd2009980

fabdrol:
@tkurki the schema validation tests kept failing, but I realised that that could be due to context. They will be added. I fixed a nagging error that caused the parser to halt after a certain number of lines. @joabakk and I are currently doing longer-term tests, before we go on with the PR.

fabdrol:
I had some time to do some profiling too, using node-server for exactly 10 minutes with the same log files (and on the same hardware) results below:

Old parser:
```
 [Summary]:
   ticks  total  nonlib   name
    343    3.8%    3.8%  JavaScript
   8262   92.4%   92.6%  C++
     63    0.7%    0.7%  GC
     20    0.2%          Shared libraries
    319    3.6%          Unaccounted
```

New parser:
```
 [Summary]:
   ticks  total  nonlib   name
    282    3.2%    3.3%  JavaScript
   8143   93.7%   93.9%  C++
     42    0.5%    0.5%  GC
     14    0.2%          Shared libraries
    251    2.9%          Unaccounted
```

In the new parser, we see reduced CPU time for the GC and javascript (which is good), and less ticks. Whilst this is only one test I would say the results are promising.

tkurki:
Please measure throughput or cpu at constant throughput. You could get those numbers with a new less efficient implementation afaik.

tkurki:
I thought I'd do a bit of timing, but the `bin/nmea0183-signalk` is generating some error messages:

```
 cat ../server-node/samples/nais300-merrimac.log | bin/nmea0183-signalk >/dev/null
no writecb in Transform class
no writecb in Transform class
no writecb in Transform class
no writecb in Transform class
no writecb in Transform class
...
```

fabdrol:
I don't think the binary has been updated, let me have a look.

fabdrol:
@tkurki that has been fixed, there was a bug where the callback of the `_transform` method was called multiple times, thanks for catching it

fabdrol:
I did some tests, suggested by @tkurki, to compare the old and new parser. Results are very close, with the new parser slightly faster. 

results

The line numbers, per parser, were as follows:

```
input.log: 294911
newparser.log: 294911
oldparser.log: 311296
```

An interesting thing I noticed is that the old parser generates more lines than it should (RMC sentences are parsed twice) and that various delta messages are not correct to specification.

fabdrol:
## Update

We've run multiple multi-day tests, and the new release is almost ready to go. There will be a few changes left: 

- [ ] Review & merge @joabakk's #92 PR
- [ ] Review multi-line AIS support
- [ ] Remove the `parser.stream()` interface - this was added as suger to ease the transition from 1.x to 2.0, but it causes a memory leak. Instead we'll add an example to the documentation on how to use it in a stream (using a wrapper). 
- [ ] Add support for proprietary NMEA0183 sentences (i.e. starting with `$P......`, such as `$PSILTBS`. See for instance [this Garmin PDF](https://www.garmin.com/support/pdf/NMEA_0183.pdf )

tkurki:
If pripietary sentence support is a new feature why bundle it in this?

joabakk:
I added it a while ago as a provision in the existing parser, but it's not in use. I'm happy to skip this for noe

fabdrol:
@tkurki I agree with you about the proprietary sentences, but I do want to make sure that they don't break the parser. I updated the todo list, all agreed?

tkurki:
Have you reviewed all the changes in master since development started?

BTW it would be a good idea to use git cherry pick where possible. For example ee673d39bb245a3440262a6ac3f061b402e62469 and ee502098ff69541c0ec6639035e301472bfcd18e.

When merging PRs I have found that rebasing or squashing (available in the Github UI) result in cleaner git history, unless the PR itself has meaningful data and you want to keep the reference in the merge commit, but your call.

Are you going to clean up the commit history? Probably not time well spent at this stage.

For the future: https://chris.beams.io/posts/git-commit/#imperative and other points on that page.

What about a624ab6 and https://github.com/fulup-bzh/GeoGate/issues/23

- [ ] use latest signalk-schema from @signalk

fabdrol:
@joabakk @tkurki I think we're ready to move on to 2.0 with this parser. Unless you guys have any objections, I'll branch master off into `1.x` and tag it with the current version, then replace master with the development branch, tag it `2.0` and release it to NPM. I also have a version of the `provider` for server that checks the installed version of the parser and uses the right code.

tkurki:
Release should include changing the server to use this.

fabdrol:
@tkurki 
> I also have a version of the provider for server that checks the installed version of the parser and uses the right code.

fabdrol:
@tkurki I went through the changes in master, apart from `navigation.datetime` we seem to be good

fabdrol:
IMHO we're ready to go.

joabakk:
No objections from me.

tkurki:
Nor from here - something holding you up?

PR #79, opened 2017-02-11
Fix #78: de-singleton ais session

### LINKED ISSUE #78: Non singleton Ais context
Currently there is only one singleton Ais context and multiple parser instances will mess up this global state. Ais context should be per parser instance.

tkurki:
Ais session has been a singleton, which will mess things up
if you run several Parsers concurrently and they have ais data.
The session is monkeypatched to multiplexer, which is the only
available context in consecutive invocations.

Fixes #78.

### DIFF
--- a/codecs/VDM.js
+++ b/codecs/VDM.js
@@ -27,10 +27,9 @@ as a global singleton. Proper implementation would be
 a per stream constructer NMEA0183 encoder that would be able
 to keep a coherent internal parsing state.
 */
-var session = {};
 
 module.exports = new Codec('VDM', function(multiplexer, input, line) {
-  var data = new Decoder(line, session);
+  var data = new Decoder(line, multiplexer.aisSession);
 
   if (!data.valid) {
     return this.reportError(this.errors.VOID, "Not parsing sentence for it isn't valid.");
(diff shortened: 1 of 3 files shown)


samothx:
OK, testing it on live data, following remarks:
Basically it seems to work. I will yet have to set up a second stream though to see how that goes. 
When a multi message sentence is processed Decoder()  returns data with data.valid set to false. Only when the message is complete data.valid is set to true. You call reportError() whenever data.valid is not true..

samothx:
More comments: ggencode is available in version 0.1.13, we are using 0.1.5.

samothx:
Might be off topic here but I am not receiving any timestamps in the delta signalk for ais vessels.. I just compared it to other delta data (for self) and there I am seeing timestamps most of the time. 
I am also finding vessels showing up in the rest api that look like this: 
"urn:mrn:imo:mmsi:512004213":{"mmsi":"512004213"}} 
Is this connected to how data is added in the Codec or is it a different issue ?

fabdrol:
This should do for now, in 2.0 version we'll test this some more. @samothx what's the use case for multiple streams? I'll add a test for that specific use case.

samothx:
@fabdrol  I do not think there is much of a use case, but it could happen if two sources of AIS are present and connected to SignalK and you don't want your AIS data messed up if it does happen.

PR #83, opened 2017-05-13
Fix return type in README

sarfata:
Script returns SignalK sparse (or full) data. Not in the "update" format.


joabakk:
#84 will fix the travis error

PR #86, opened 2017-07-01
correct timestamp RMC

e-sailing:
Month for January is 0 not 1. So it was one month to high.

### DIFF
--- a/lib/NMEA0183.js
+++ b/lib/NMEA0183.js
@@ -181,7 +181,7 @@
     if (date) {
       var year, month, day;
       day = this.int(date.slice(0, 2), true);
-      month = this.int(date.slice(2, 4), true);
+      month = this.int(date.slice(2, 4)-1, true);
       year = this.int(date.slice(-2));
 
       // HACK copied from jamesp/node-nmea


joabakk:
I'm not so sure. My normal sources do not mention this. Do you have a source for this?

e-sailing:
When I search for the right date time. I saw this timestamp difference.
{
	u'source': {
		u'label': u'OPkplex'
	},
	u'values': [{
		u'path': u'navigation.position',
		u'value': {
			u'latitude': 52.70580083333333,
			u'source': {
				u'label': u'signalk-parser-nmea0183',
				u'type': u'NMEA0183',
				u'talker': u'GP',
				u'sentence': u'RMC'
			},
			u'longitude': 8.664118666666666,
			u'timestamp': u'2017-08-01T18: 56: 13.000Z'
		}
	}],
	u'timestamp': u'2017-07-01T18: 56: 13.057Z'
}

Then I googled this.
https://stackoverflow.com/questions/8335276/javascript-getutcmonth-returns-0-for-december
After the change the timestamp was OK.

joabakk:
Understood. So it's an implementation thing. That probably all month instances from NMEA are the same.

PR #90, opened 2017-07-07
add navigation.datetime

e-sailing:
navigation.datetime does work on NMEA2000 (pgn 126992) but not on NMEA0183
RMC has the complete date and time. Why keep differences between Signal K keys from 2000 to 0183?

### DIFF
--- a/codecs/RMC.js
+++ b/codecs/RMC.js
@@ -63,38 +63,39 @@ module.exports = new Codec('RMC', function(multiplexer, input, line) {
 
 	var ts 	 = this.timestamp(time, date);
 	var self = this;
-
-  // Position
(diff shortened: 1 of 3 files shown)


tkurki:
Could you please add a test for RMC, preferably with a sample from real world. See other tests in the test directory for examples.

e-sailing:
How do I start an example from test directory?

tkurki:
npm test

runs all the tests.

tkurki:
Merged manually, since I fixed the commit messages (N2K parser's features are not really relevant, just the fact that navigation.datetime is added) and improved the test messages a bit.

Thank you!

PR #98, opened 2017-10-17
fix: HDG with just heading

tkurki:
The new parser is not working with HDG with just the heading field, which my equipment happens to produce.

### DIFF
--- a/hooks/HDG.js
+++ b/hooks/HDG.js
@@ -39,7 +39,20 @@ module.exports = function (parser, input) {
   try {
     const { id, sentence, parts, tags } = input
 
-    if (isEmpty(parts[0]) || isEmpty(parts[3]) || isEmpty(parts[4])) {
+    const values = []
+    if (!isEmpty(parts[0])) {
+      values.push({
+        path: 'navigation.headingMagnetic',
+        value: utils.transform(utils.float(parts[0]), 'deg', 'rad')
+      })
+    }
+    if (!(isEmpty(parts[3]) || isEmpty(parts[4]))) {
+      values.push({
+        path: 'navigation.magneticVariation',
+        value: utils.transform(utils.float(parts[3]), 'deg', 'rad') * (parts[4] === 'E' ? 1 : -1)
+      })
+    }
+    if (!values.length) {
       return Promise.resolve(null)
     }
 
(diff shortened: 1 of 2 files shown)


joabakk:
That one seems to come from d53aefe09f87cc45093ac240bb4268af5fc87819. Looks like it should have been ands and not ors? We introduced tests for empty sentences, and this one is wrongly reported as such.

tkurki:
It is not just the check, you need to alter the way the values array is constructed.

joabakk:
I see a few issues:
APB is incomplete
DBT empty sentence test logic (DSC type?)
GGA should accept more than 3 empty fields?
GLL, does it choke on first empty field?
HDG as mentioned
MTW, RMC, RPM, VDR, VHW missing empty test logic
VWR should accept more than 3 empty fields?

So quite a list, but still it should have been discovered before

fabdrol:
Hmm good find, I didn’t realise that was considered a valid case (the test for choking on empty sentences verifies that the thing returns no data in that case). 

@tkurki: could you send me those sentences? I’ll change the parser’s behaviour

tkurki:
What sentences? Did you check the code in this PR?

fabdrol:
Not yet, I was on a mobile device. But I'll go through it for every hook. A question I'd like to ask is: is there one generic way to assess the invalidity of a incoming sentence, apart from (1) it being empty and (2) an invalid checksum? I'm proposing to add the field check to the parser before a hook is invoked (i.e. to check if it's empty) and defer any other validity check to the hooks. The hook can then take care of checking VOID flags in certain sentences, validating fields in others..

tkurki:
Why are you adding more validation - I have not heard of any problems related to this with the old parser, is there something to fix?

fabdrol:
I'm not saying I'm adding more validation, I was saying I'd move validating that a sentence is not empty out of the hooks and _asking_ if there're any more validations we need to do.

tkurki:
What do you mean by _a sentence is not empty_? That it is not an empty string? Or that all fields are empty? If it is an empty string it won't be converted anyway.

My point is that for example HDG the hook can convert the data that is there and not worry about any other validation. 

In some cases there needs to be data in several fields, like variation and deviation need both the magnitude and direction. In a sense that is validation: _only output variation if both magnitude and direction fields have valid values_.

This is what it did previously: https://github.com/SignalK/signalk-parser-nmea0183/blob/v0.6.0/codecs/HDG.js#L50-L56. The conversion could be a but more strict, it only checks for `E` and `something else, but truthy`.

tkurki:
I would really love to have a way to log everything that is **not converted** for testing and debugging, now during this upgrade and later in prod use when you get questions about _why is such and such not being shown_.

joabakk:
I added a second debug variable and called that if the sentence was not converted. It only returned empty sentences and those without hooks for the plaka file, as could be expected. But we should not limit ourselves to the few n0183 logfiles in the server install.

joabakk:
Good idea to check if all fields are empty before sending it to the hook @fabdrol. Makes it cleaner and easier to add new hooks. Then there are more exotic variants like HDT sending "here is the heading, but it's not valid" that need to rest in the hook. Then we could add a `nmea0183-parser-skipped` or similar debug message with the dropped sentences for the user to troubleshoot

PR #116, opened 2018-03-26
feature: identify proprietary sentences with full name

joabakk:
Using the existing rules for proprietary sentences, the `talker` would be "P*" and the `id` would be the three next letters. Say for `$PNKEP`, the `id` of `KEP` could be used by other sentences and cause confusion as there are no `$**KEP` sentences in the "official" sources.

### DIFF
--- a/hooks/PNKEP.js
+++ b/hooks/PNKEP.js
@@ -1,4 +1,4 @@
-"use strict"
+"use strict";
 
 /**
  * Copyright 2018 Signal K <info@signalk.org> and contributors.
@@ -16,8 +16,8 @@
  * limitations under the License.
  */
 
-const debug = require("debug")("signalk-parser-nmea0183/KEP")
-const utils = require("@signalk/nmea0183-utilities")
+const debug = require("debug")("signalk-parser-nmea0183/PNKEP");
+const utils = require("@signalk/nmea0183-utilities");
 
 /*
 === PNKEP,01 - NKE Target speed ===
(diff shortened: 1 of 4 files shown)


test/PNKEP.js
```diff
@@ -36,7 +36,7 @@ describe("KEP", () => {
         4.269889970594349,
         0.0005
       )
-      toFull(delta).should.be.validSignalK
+      //toFull(delta).should.be.validSignalK
```
sarfata: why did you comment this?

hooks/PNKEP.js
```diff
@@ -62,33 +62,25 @@ Field Number:
...
-    (typeof mixed === "string" && mixed.trim() === "")
-  )
-}
-
 module.exports = function(parser, input) {
   try {
-    const { id, sentence, parts, tags } = input
-
-    var delta
+    const { id, sentence, parts, tags } = input;
```
sarfata: A lot of the changes in this file are about adding ';' to end of lines. I did not do that in my other PR because it looked like that was not the style chosen by the original author. 

Is there a SignalK "coding guidelines" document somewhere?

test/PNKEP.js
joabakk: I honestly don’t remember, perhaps it was a lazy fix to a failing test

joabakk:
I have not come across this, but I find it cleaner to separate out the proprietary ones.

test/PNKEP.js
tkurki: If you replace the commented line with 
```
      try {toFull(delta).should.be.validSignalK} catch (err) {console.log(err);done(err); return}
```
you get
```
...
/sources/DUMMY_LABEL/PNKE/sentences/PNKEP:Unknown property (not in schema)
```
I'd like to move forward with the Promise removal, so how do we move forward?

joabakk: Ok. I’ll close this an try again after promise removal

PR #117, opened 2018-04-08
seconds are missing

e-sailing:
The timestamp function in nmea0183-utilities has the issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "author": "Fabian Tollenaar <fabian@signalk.org> (http://signalk.org)",
   "license": "Apache-2.0",
   "dependencies": {
-    "@signalk/nmea0183-utilities": "^0.5.0",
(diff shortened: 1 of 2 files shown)


sbender9:
Now the test is failing because of this change

sbender9:
Can you update the nmea0183-utilities version to 0.6.0? Then the test should pass and I can merge

sbender9:
Sorry, I'm asking you to update the PR. Please update the utilities version and the tests, We won't merge a pull requests with the tests failing.

e-sailing:
I tried to fix it  with another pull request #119 (0.5.0 to 0.6.0) . But it has the same problem the other way round. The result is now with the 12 seconds but this part  with the 12 seconds isn't merged and gives an error. Please help.

sbender9:
Put the version change in this pull request.

PR #121, opened 2018-05-12
style: Enable Eslint

webmasterkai:
I've started on some aslant with fixes. Maybe someone should let me know if I'm headed in the right direction with this.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,15 @@
+---
+env:
+  browser: false
+extends:
+  - eslint:recommended
+  - airbnb-base
+  - plugin:import/errors
+  - plugin:import/warnings
+plugins:
+  - import
+rules:
+  semi: [2, never]
+  strict: [0, global]
+  no-underscore-dangle: 1
+  import/no-dynamic-require: 1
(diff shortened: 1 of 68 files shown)


tkurki:
No.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get
https://stackoverflow.com/questions/31605391/how-does-the-chai-expect-function-work

webmasterkai:
I like that solution. Nice find.

Should I continue on fixing the errors?

tkurki:
Some tests are failing, any idea why?

Sure, go ahead with the fixin’.

tkurki:
And how do you hook eslint so that CI tests conformance?

tkurki:
Any plans to push this forward? This touches gazillion files and will get out of date with #125.

webmasterkai:
This is too large a project for me to enable and fix all errors. It's just more time than I have at the moment.

How do you feel about establishing some rules and then implementing them in smaller chunks and with every new commit?

tkurki:
Adding just 1 rule and setting up eslint in general so that contributors can easily use it would be valuable.

PR #122, opened 2018-05-23
Fix: VWR, ok if 4 empty fields (eg. fields 4-7)

davidsanner:
Example sentence:  $IIVWR,024,L,018,N,,,,*5e   should be accepted (w/ 4 missing values)
var 'empty' not > 4 then continue / set environment.wind values.

### DIFF
--- a/hooks/VWR.js
+++ b/hooks/VWR.js
@@ -41,7 +41,7 @@ module.exports = function (parser, input) {
   const { id, sentence, parts, tags } = input
 
   const empty = parts.reduce((count, part) => { count += (isEmpty(part) ? 1 : 0); return count; }, 0)
-  if (empty > 3) {
+  if (empty > 4) {
     return Promise.resolve(null)
   }
 
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -39,10 +39,8 @@ module.exports = function parseSentence(emitter, sentence, options) {
 
   let validateChecksum = _.isUndefined(options) ||
       _.isUndefined(options.validateChecksum)  || options.validateChecksum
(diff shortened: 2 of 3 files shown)


tkurki:
Would you be able to add a test for this?

davidsanner:
Is this the test you are looking for?   The test is a valid sentence with only 4 values set which was failing before:  '$IIVWR,024,L,018,N,,,,*5e' 

On a side note there are older instruments out there that do not append a checksum to their nmea 0183 sentences.  I made some changes to NMEA0183-utilities that would still pass/fail sentences with checksums but pass any sentence without an '*' in it.  I can do a pull request if you like though I thought it might be better if there was an option in the setup screen which I haven't dug through enough to do yet.

sbender9:
@davidsanner There is an option to disable checksum validation for this reason, but I think there should also be an option to disable the check only if '*' is missing like you have done. Please do the pull request, I can do the UI part.

sbender9:
This PR looks good to me. @tkurki ?

sbender9:
David, pleas don’t mix unrelated changes in the same PR. Please remove this last commit and create a new PR for the change.

davidsanner:
Sorry about the unrelated commits in single PR.  That's what I get for messing around in the master branch.  PR reverted to simple/minimal bug fix.  Added separate PR here https://github.com/SignalK/signalk-parser-nmea0183/pull/123

sbender9:
Looks good, thanks!

davidsanner:
Still getting up to speed on testing - should be good to go now.  Thx.

tkurki:
Published in v2.3.1

PR #123, opened 2018-05-31
feature: add option to accept without checksum

davidsanner:


davidsanner:
This PR should go in after https://github.com/SignalK/nmea0183-utilities/pull/9

tkurki:
What would be the use case where you want to accept sentences only if `*` is present, but ignore the actual checksum? Or want to check for `*` but ignore the checksum? 

What I'm after is that wouldn't it be simpler to modify the `valid(...)` function so that if `validateChecksum` is false it does not check for `*`?

Having two options that are almosta the same, one of which is `true` and the other is `false` in the lenient case, seems needlessly complex to me.

tkurki:
What do you mean with `enumerate validateChecksum`?

See also https://github.com/SignalK/signalk-server-node/issues/549

As nmea0183 is mostly one talker, multiple listeners you won't bump into the _some instruments don't generate checksums at all, some do_ case without a multiplexer. If you have multiple 0183 inputs on the SK server you can configure validateChecksums per input. All this sounds like the simple solution of allowing missing * when validateChecksums = false would be my choice. That can be done by just a change in utils.

davidsanner:
I have 4 inputs going through kplex first as that was my original setup and initially only used SignalK to route 10Hz GPS & heading from n2k into kplex.  (RMC sentences are still jumping between local and UTC timezones).   Now that I've jumped more into SignalK I'm routing my 0183 back into signalK via kplex though the filters are getting complex to avoid looping.  Perhaps it's time turn it all around and send 0183 into signalk first though signalk-to-nmea0183 doesn't support all the sentences coming out I'm generating and I'm not sure when RMC is generated which GPS source it will use... both?

I digress...  agreed, typical use case setting validateChecksums = false on a per input configuration seems like they way to go and no need to add complexity with this PR.   Feel free to close this PR and https://github.com/SignalK/nmea0183-utilities/pull/9

PR #125, opened 2018-07-30
Convert Promises to a synchronous api to make calling parse simpler

sarfata:
The use of promises inside the parser-nmea0183 library forces caller into an asynchronous mode but we do not actually need Promises here since all the data can be parsed/calculated immediately.

This PR should be completely backwards compatible for callers. The only difference will be that the event will be emitted immediately when calling the `parse()` function, instead of being called during the next tick. Hopefully this will not impact anyone.

I have tried to standardize the hooks so they all look similar which added a few "noisy" changes.
Conventions I used:
 - Hooks return the `delta` object directly `return {}` instead of `const delta={}; return delta;`
 - Hooks do not catch exceptions. They are caught in the `Parser.parse()` function instead

I hope you will find this change acceptable. I think it will make this library much more useful in context where going asynchronous is complicated or would impact performance (like reading files with 1000s of NMEA lines).


hooks/ROT.js
```diff
@@ -36,28 +36,23 @@ module.exports = function (parser, input) {
   const { id, sentence, parts, tags } = input
 
   if (String(parts[1]).toUpperCase() !== 'A') {
-    // Don't parse this sentence as it's void.
-    return Promise.reject(new Error('Not parsing sentence as data is not valid)'))
+    return null
```
sarfata: I have changed the behavior here because I think this is more coherent with what we do in other hooks. Happy to discuss / change everything to throw instead of returning null.

test/ALK.js
```diff
@@ -119,16 +120,15 @@ describe('ALK', done => {
...
-  it('Doesn\'t choke on empty sentences', done => {
-    const parser = new Parser
-    parser
-    .parse('$STALK,9C,,,*3B')
-    .then(result => {
-      should.equal(result, null)
-      done()
-    })
-    .catch(e => done(e))
+  it('Doesn\'t choke on empty 0x9C sentences', () => {
```
sarfata: Re-enabled this test and added a similar one for 0x84. Changed sub-hooks to make them pass.

test/GGA.js
```diff
@@ -26,11 +26,11 @@ chai.use(require('@signalk/signalk-schema').chaiModule)
...
 
 describe('GGA', () => {
-
   it('Converts OK using individual parser', done => {
-    const parser = new Parser
+    const parser = new Parser()
 
     parser.on('signalk:delta', delta => {
+      // FIXME: label is a required property in the SK Spec.
       should.not.exist(delta.updates[0].source.label)
```
sarfata: Anyone remembers why we are testing **against** the presence of a source label? A source label is required in the spec.

lib/parse.js
```diff
@@ -44,7 +44,9 @@ module.exports = function parseSentence(emitter, sentence, options) {
 
   if (valid === false) {
     emitter.emit('warning', { message: `Sentence "${sentence.trim()}" is invalid` })
-    return Promise.reject(new Error(`Sentence "${sentence.trim()}" is invalid`))
+    // FIXME: Do we really want an exception (or Promise.reject) here?
```
sarfata: Comments welcomed here: Should we return null or throw an exception when the checksum is invalid?

sarfata:
@fabdrol  / @tkurki / @joabakk - I would love to know how you feel about this big PR. Let me know if you have any concerns.

tkurki:
I have just read the title, but I'm all for it. 

For example log playback to insert timestamped data into a db (with backpressure) doesn't really work with async, currently timestamps get mixed up.

joabakk:
No objections from me, but I'd like to see #116 merged as well

tkurki:
Running `time bin/nmea0183-signalk` with the JSON output commented out on `plaka.log` from server's samples I get about 950ms with the master version and about 770ms with this version, so not too bad.

@sarfata just curious - where do you use this? I assume you did not do it just for kicks ;-)

tkurki:
@sarfata could you squash the commits? And I do appreciate commit messages in the imperative mood: https://chris.beams.io/posts/git-commit/#imperative

.vscode/launch.json
```diff
@@ -0,0 +1,33 @@
+{
```
tkurki: Accidental addition?

hooks/DBT.js
```diff
@@ -37,31 +37,26 @@ Field Number:
...
-              path: 'environment.depth.belowTransducer',
-              value: utils.float(parts[2])
-            }
-          ]
-        }
-      ],
-    }
-    
-    return Promise.resolve({ delta })
-  } catch (e) {
```
tkurki: ❤️ getting rid of the trycatches in all decoders.

.vscode/launch.json
sarfata: No. This makes it possible to run the unit test and debug them very easily in VSCode. I thought other people might enjoy it. Should work without any change on anyone's computer.

I would also completely understand if you would rather keep the repo IDE-agnostic ;)

sarfata:
> could you squash the commits?

I can do another PR with commit squashed. Is this what you meant? Otherwise, I believe you can squash and edit the commit message when you merge the PR. I don't think I can do this.

In any case, what do you think of this commit message:
```
Refactor library to not use Promises internally

The use of promises inside the parser-nmea0183 library forces caller into an 
asynchronous mode but we do not actually need Promises here since all the data 
can be parsed/calculated immediately.

This PR should be completely backwards compatible for callers. The only 
difference will be that the event will be emitted immediately when calling the 
parse() function, instead of being called during the next tick. Hopefully this 
will not impact anyone.
```

> where do you use this? I assume you did not do it just for kicks ;-)

I am working on a browser based log visualizer that can load logs in any "common" format.  Of course SignalK logs is one of them (that is what KBox generates too now). The strategy is to use a loader that converts in SignalK, adjusts the sampling and then displays it. v0 is almost ready to share, just working on some performance issues now so it's more sexy ;)

I had a few more changes I needed to do to the library (and canboatjs/signalk-nmea2000 too) to be able to use it with webpack in the browser. The biggest is getting rid of the "automatic" loading of supported sentences by looking at the filesystem.

See: https://github.com/sarfata/signalk-parser-nmea0183/commits/feature/webpack-compatible

I will push more PRs once this one is merged and you can all tell me if hardcoding the list of sentences in a JS file is acceptable, in exchange of being able to run in the browser - or maybe you will think of another way to go around that issue.

sarfata:
Can someone more familiar with signalk-server than me verify that this PR does not break it?

tkurki:
Sure you can squash the commits and thne force push to the same branch, that will update this pr automatically. https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

Commit message is good. I would mention the new immediate function though and not talk about a pr.

A hardcoded sentence list is ok. It is not like sentences come and go all the time.

The server has code that autodetects log format - we could have extracted that, or maybe can in the future. 

In browser log playback is something that I’ve been thinking about, looking forward to what you have come up with!

I can check server compatibility.

tkurki:
As commits go .vscode is not related to te Promise removal and should go in a separate commit, if not even a separate PR. Like if we start producing release notes from prs.

test/GGA.js
tkurki: The parser does not know what the label of the source should be - it depends on server configuration and the server will overwrite it.

lib/parse.js
tkurki: null

lib/index.js
```diff
@@ -40,18 +38,33 @@ class Parser extends EventEmitter {
...
+      var result = parseSentence(this, sentence, this.options)
+      if (typeof result === 'object' && result !== null) {
+        this.emit('signalk:delta', result || {})
+      }
+      // Return value kept for backwards compatibility. The usage of
+      // parseImmediate("") is recommended instead.
+      return Promise.resolve({ delta: result })
+    }
+    catch(e) {
+      // FIXME: Is this really useful? Maybe should just throw the exception
```
tkurki: To be backwards compatible this should return a rejected Promise?

tkurki: Returning null is sufficient here I think - are we throwing errors somewhere?

sarfata:
Commits squashed and I also updated the README to mention the new API.

However, before we merge I have a few extra concerns with the new API that I think we should address before we release a version including it. Otherwise we will have the burden of backwards compatibility in the future.

They are:
1. Right now, in case of error, the new `parseImmediate` API sometimes return null and sometimes throws an exception. 
   * I propose always throwing something when we cannot return a delta object. This way the caller does not have to check if the return value is null and if there is an error, it's always explicit what the error is (wrong checksum, unknown sentence, etc).

2. The API require creating a parser object with `new Parser()` before calling `parseImmediate`. There really is no solid reason for this. We never use the "context" (parser.session) as far as I can tell.
   * I propose exporting a `parseNMEA0183(sentence, options = {})` function directly.

3. The export is not ideal in an es6 world because it's exporting the class directly (`module.exports = Parser`). I think it would be better to have `module.exports.default = Parser`, this way we could import with `import Parser from '@signalk/nmea-0183-parser'` in ES6 environments. 

    As it is now, it's possible to make it work with some TypeScript or Babel options (see [`esModuleInterop` for example](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-form-commonjs-modules-with---esmoduleinterop) but we should improve this.

As I am writing this, I am realizing that this `parseImmediate()` function requires more discussion and is not ready. The changes I am listing here might be more breaking so I am going to remove `parseImmediate` from this PR, re-squash and we can merge this PR and discuss the new API in another one.
(started working on removing parseImmediate tonight but some tests use it now so it's going to take some effort)

lib/index.js
sarfata: > To be backwards compatible this should return a rejected Promise?
> Returning null is sufficient here I think - are we throwing errors somewhere?

I saw that tonight too while trying to remove some changes and simplify this PR. It's more complicated than I thought to be really backwards compatible.

As far as I can tell, current code throws exception in a few places that are converted in `emit('error')` [in `lib/index.js`](https://github.com/SignalK/signalk-parser-nmea0183/blob/master/lib/index.js#L53):

```
  parser
    .parse(data)
    .then(result => {
      if (result != null) {
        console.log(JSON.stringify(result.delta))
      }
    })
    .catch(e => {
      console.error('Encountered an error:', e.message)
    })
```

In `bin/nmea0183-signalk` we have:
```
    return parseSentence(this, sentence, this.options)
      .then(result => {
        if (typeof result === 'object' && result !== null) {
          // this.emit('signalk:full', result.full || {})
          this.emit('signalk:delta', result.delta || {})
        }
        return result
      })
      .catch(e => {
        this.emit('error', e)
      })
```

The `promise.catch` is never reached because we already have a catch in `lib/index.js`. Instead, the `promise.then()` of `bin/nmea0183-signalk` is called with whatever was returned from the `catch` statement, in this case we have no return statement so the catch returns `undefined`.

So to be backwards compatible, we need to:
 - `emit(error)`
 - Return a `Promise.resolve(undefined)`

Let me know if you disagree. I will update the code to match this.

tkurki:
Multiline AIS VDM messages use the session: https://github.com/SignalK/signalk-parser-nmea0183/blob/master/hooks/VDM.js#L54.

Your plan sounds good.

lib/index.js
tkurki: Well, the cli utility is internal to this module, so we don't need to pay attention to being backwards compatible there.

Being backwards compatible is (was?) more about being possibly more confident that this not cause regression errors than supporting a multitude of projects that rely on current functionality.

I sincerely doubt anything but your project and sk server use this to any meaningful extent. We can break backwards compatibility and adjust the server's code accordingly if we want to. In fact to get real benefit from the synchronous api to fix the timestamp related problem I mentioned earlier we should adjust the server's code.

So let's make the new api good with less convoluted error handling (maybe a test or two for that) and not worry too much.

test/GGA.js
sarfata: I think the library should always return valid SignalK. Any objection to me fixing this in another PR? The server can override it later.

tkurki: Always returning valid Signal K makes sense. We don't need to override it - you can pass the needed data in Parser constructor.

sarfata:
Closing this. #130 has a much better commit-log and some small improvements.

PR #126, opened 2018-08-07
feature: Parse VLW

jtroyer76:
Parse VLW nmea sentences.  Total cumulative distance posted to
navigation.log.  Distance since reset posted to navigation.trip.log


tkurki:
Published as version 2.4.0 in npm. Thanks!

PR #127, opened 2018-08-13
refactor: do not use Promises internally

sarfata:
New version of the PR to remove promises (internally only for now):
 - first commit adds a test for APB sentences (they always throw an error). This was helpful to make sure I did not break error handling before/after the other commit.
 - second commit is the big change, it removes try/catch and promises from all the hooks and replaces them by one try/catch in `index.js`'s `Parser::parse()`.

The tests have not been touched (except to re-enable one that was disabled) so it should be a lot easier to review this PR and make sure it does not change more than what it pretends to change. 

I will add more PR to update the API and make sure we do not return a promise to the caller and always handler errors in the same way. The other PR will be more complicated but easier to review since it will not include all the hooks changes we have here.

### DIFF
--- a/test/APB.js
+++ b/test/APB.js
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2016 Signal K and contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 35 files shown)


test/APB.js
```diff
@@ -0,0 +1,48 @@
...
+const chai = require('chai')
+const should = require('chai').Should()
+
+const heading = '$STALK,84,B6,10,00,00,00,00,00,00*14'
+const standby = '$STALK,84,E6,15,00,00,00,00,00,08*1E'
+const auto = '$STALK,84,56,5E,79,02,00,00,00,08*16'
+const wind = '$STALK,84,06,00,00,04,00,00,00,00*63'
+const route = '$STALK,84,06,00,00,08,00,00,00,00*6F'
+const rudder = '$STALK,84,06,00,00,08,00,FE,00,00*6C'
+const heading_nineC = '$STALK,9C,51,1E,00*4B'
```
tkurki: Unuses lines?

sarfata: Force-pushing a fix.

PR #130, opened 2018-08-16
feature: new immediate-style API to replace Promise/events  …

sarfata:
Collection of changes that lead to:
- a new API that should be easier to use
- compatibility with webpack

Biggest change is the last one:
```
Replace the mix of promises and events in the old API by a simpler API
that returns a SignalK Delta object, null if there was no valid data
(invalid gps fix, unsupported sentence or partial AIS message for
example) and throws exception when something goes wrong (invalid
checksum, invalid data fields, etc).

Not using promises makes it easier for callers to use the library and
should result in less memory usage and slightly improve performance.
```

We need to merge https://github.com/SignalK/specification/pull/498 and update the dependency before we are really compatible with webpack.


lib/index.js
```diff
@@ -33,28 +36,54 @@ class Parser extends EventEmitter {
     this.license = pkg.license
   }
 
-  parse (sentence) {
-    if (typeof sentence === 'string' && sentence.trim().length > 0) {
-      this.emit('nmea0183', sentence.trim())
```
tkurki: This is something we want to keep: emitting the raw sentences. These are available within the server for plugins and for example over tcp.

sarfata: Would it be hard to emit the sentences from the same place where you call `parser.parse()`? 

I can re-introduce it but it seems pretty orthogonal to the library itself. There is absolutely no value added besides `.trim()`.

tkurki:
Are we done here? Could you take care of changes in the server next?

sarfata:
@tkurki I am looking into the server now. Can you publish version 2.5.0 of the package to npm?

PR #132, opened 2018-08-29
SeaTalk sub hooks

raffmont:
Added some new STALK sub-hooks.


joabakk:
Could you provide tests for the new datagrams? They can be added in test/ALK.js

tkurki:
Superceded by #135

PR #133, opened 2018-08-31
feature: add an option to ignore missing checksums

sarfata:
This does not change the default behaviour: missing checksums will be
treated as invalid checksums.

However, with this new option it's possible to accept missing checksums,
without accepting sentences that have invalid checksums.

Requires: https://github.com/SignalK/nmea0183-utilities/pull/10/files


tkurki:
Is still valid? To me this looks like something that might be usable.

PR #134, opened 2018-09-07
feature: identify proprietary sentences with full name

joabakk:
Re-opening #116 after recent changes to repo.


tkurki:
Schema specifies sentences property name as three letters, you have `PNKEP`.

https://github.com/SignalK/specification/blob/master/schemas/groups/sources.json#L37

joabakk:
Thanks @tkurki it should be ok now

joabakk:
Can we merge this?

joabakk:
Squashed into a single commit and force pushed. Third time lucky...

tkurki:
Are the changes in VWR related?

Prefer let/const over var, please change.

joabakk:
The VWR tesst used PIVWR, which in NMEA0183 book identifies it as a proprietary sentence, that's why it had to be changed. Changed to const in lib/index

PR #135, opened 2018-09-07
Add seatalk datagrams with test

joabakk:
Tests have been produced backwards, so the calculations may still be off. Real world data and verification is welcomed


joabakk:
Also, the different timedate datagrams are sketchy. If one datagram sends time and another date, is it correct to assume today's date or time to fill in `navigation.datetime`?

joabakk:
This is meant to replace #132

PR #142, opened 2019-02-12
feature:parse PBVE (oil pressure only)

tonybentley:
Oil pressure from Cruz Pro proprietary sentence

FYI Cruz Pro has several instruments that use the proprietary sentence PBVE so I am supporting only one sentence type (oil pressure). Will later abstract each PBVE sentence type for modularity. 

Added unit test that checks if the productCode is not 'D', expecting undefined. Not sure if this is okay. Let me know.

screen shot 2019-02-12 at 11 38 56 am


rob42:
FYI, I created a mapper for this a several years ago in the old signalk-java-server. The comments and format data may help you for the other fields. 

See https://github.com/SignalK/signalk-server-java/blob/master/src/main/java/net/sf/marineapi/nmea/sentence/BVESentence.java

joabakk:
There should be an instance of propulsion before `oilPressure`. How about adding the full message decoding in the same PR?

README.md
```diff
@@ -10,6 +10,7 @@
 
 - [ALK - Seatalk](https://en.wikipedia.org/wiki/Seatalk)
 - [APB - Autopilot Sentence "B"](http://www.catb.org/gpsd/NMEA.html#_apb_autopilot_sentence_b)
+- [BVE - CruzPro Proprietary Sentence (currenly only OP30/60 supported)](http://www.cruzpro.com/op60.html)
```
tkurki: typo in currently

tonybentley:
> There should be an instance of propulsion before `oilPressure`. How about adding the full message decoding in the same PR?

So I will hard code propulsion.0.oilPressure in the 

> There should be an instance of propulsion before `oilPressure`. How about adding the full message decoding in the same PR?

So the path will be propulsion.0.oilPressure? I am uncertain how to provide the other fields for SignalK. Is there an example of how to add threshold values, etc? I can certainly add everything provided, but I'm not sure how to do that. There are a few fields that are static, and the remaining can be modified by the instrument. How would that look in code?

tonybentley:
> Could you please add a test taht checks that the output is valid Signal K - that will validate also the point by @joabakk, having the propulsion instance in the path.
> 
> See:
> 
> * [signalk-parser-nmea0183/test/VDR.js](https://github.com/SignalK/signalk-parser-nmea0183/blob/59d493417c8eff761bd475711f1d69702b4ece7b/test/VDR.js#L39)
>       
>       
>            Line 39
>         in
>         [59d4934](/SignalK/signalk-parser-nmea0183/commit/59d493417c8eff761bd475711f1d69702b4ece7b)
>       
>       
>       
>       
>   
>           
>             
>              toFull(delta).should.be.validSignalK
> * https://github.com/SignalK/specification/blob/3ccbd2a904698c6479289a0202311da591f04a68/samples/delta/docs-data_model.json#L14

done! Thanks that helps a lot

README.md
tonybentley: done

tonybentley:
> There should be an instance of propulsion before `oilPressure`. How about adding the full message decoding in the same PR?

okay the instance was added

tkurki:
Thanks!

PR #144, opened 2019-02-19
feature: PBVE -  auxillary engine temperature

tonybentley:
Added product code E, engine coolant temperature
Added the following meta for providing CruzPro gauge information:

- Backlight level
- Gauge units
- High alarm value, included in zones with messaging
- Low alarm value, included in zones with messaging
- Gauge alarm armed
- Original value for debugging

Added conversion into signalk specified format (f->k, psi->pa)
Updated unit tests to verify all meta model values

Verified with @sk/instrumentPanel and CruzPro gauges (see attached photos)
Note: Temperature is off one degree due to latency between photo and screenshot. Numbers are identical in realtime.

skinstrument

img_7797

### DIFF
--- a/hooks/proprietary/PBVE.js
+++ b/hooks/proprietary/PBVE.js
@@ -20,10 +20,15 @@
 
 const debug = require('debug')('signalk-parser-nmea0183/PBVE')
 const utils = require('@signalk/nmea0183-utilities')
+const alpha = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('')
+const schema = {
 
 /*
 
-OP30 digital oil pressure gauge sample NMEA sentence:
+Note: oil pressure calculation formulas are the same for all instruments
+
+
+OP30/OP60 digital oil pressure gauge sample NMEA sentence:
 
 Sentence: $PBVE,DGOIADNNACAEACAAABBLAAEBAACMCFAAEPAIKI*37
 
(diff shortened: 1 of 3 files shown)


joabakk:
Looks like solid work, at least in proving the accuracy!

hooks/proprietary/PBVE.js
```diff
@@ -34,85 +39,229 @@ $PBVE,x,x,xxxx,xxxx,xx,xx,xx,xx,xxxx,xxxx,xx,xxxx,xxxx,xx*xx
...
+ADAB = 16*A + D + 4096*A + 256*B
+ADAB = 16*0 + 3 + 4096*0 +256*1
+ADAB = 0 + 3 + 0 + 256
+ADAB = 259 deg F
+
 */
-const ALPHA = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('')
+  E: {
+    path: 'propulsion.0.coolantTemperature',
+    meta: {
```
tkurki: I don't think this will accomplish what you are after here. Check the full that results from this - does it have the meta in place that should be there? 

We don't have a way to update meta information with a delta. @sbender9 was working on this at https://github.com/SignalK/signalk-server-node/pull/481 but that stalled. This use case could be the time to resurrect that line of work.

tkurki:
Even if the meta stuff is off I am going to merge this, as the conversion itself works.

For the meta values it would be useful to not send them all the time, only when they change. And as I commented earlier the downstream support for updating meta is not there yet.

PR #147, opened 2019-04-20
Adds support for NMEA 0183 MDA sentence type

thorsteinssonh:
Adds support for NMEA 0183 MDA sentence type, containing meteorological composite.
This message type is used by Thies Clima, and probably other major instrument makers.

Changes to be committed:
	modified:   README.md
	new file:   hooks/MDA.js
	modified:   hooks/index.js
	new file:   test/MDA.js

### DIFF
--- a/hooks/MDA.js
+++ b/hooks/MDA.js
@@ -1,0 +1,104 @@
+/**
+ * Copyright 2016 Signal K <info@signalk.org> and contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+ const utils = require('@signalk/nmea0183-utilities')
+
+ /*
+  * MDA - Meteorological Composite
+  *
+  *        1     2     3     4     5   6   7     8     9     10    11    12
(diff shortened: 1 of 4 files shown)


tkurki:
Thanks! This is mergeable, but I suggest two changes.

You could make it a little bit more elegant / less verbose if you construct the `values` array directly as you go through the fields instead of setting variables conditionally and then later on having another null condition check on what you just set a few lines earlier.

Like this:

```
  const values = []
  ...
  if( parts[4] != '' ) {
    values.push({
      path: 'environment.outside.temperature',
      value: utils.transform(utils.float(parts[4]), 'c', 'k')
    })
  }
  ...
  const delta = {
    updates: [
      {
        source: tags.source,
        timestamp: tags.timestamp,
        values
      }
    ],
  }
```

Another nitpick would be to change `!=` to `!==`,  which is a good practice to do always in Javascript, even if it makes no difference here. See https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

thorsteinssonh:
Thanks tkurki and sbender9.
Ok I have pushed the improvements suggested by tkurki.
Note that this MDA message type can result in duplicate outside.pressure and wind.speedOverGround, depending on how the manufacturer uses the message.

It seems best just to allow those duplicates, if the manufacturer puts that into the message, even though they should contain exact same values.

Reason for the duplicates is that MDA definition is rather clumbsy, allowing wind speed and pressure entries in differen units. In signalk this just gets converted to SI units.

hooks/MDA.js
```diff
@@ -0,0 +1,127 @@
...
+  *  12.    Checksum
+  */
+
+module.exports = function (input) {
+  const { id, sentence, parts, tags } = input
+
+  const values = []
+
+  // make SI units override any non-SI units
+  let pressure = null
```
tkurki: Unused variable here

thorsteinssonh: oopps, fixed.
I should use a linter.

tkurki:
Well, we really should have a precommit hook to run lint.

Thank you very much!

PR #152, opened 2019-07-05
PASHR NMEA Sentence

yyvus:
Communicates states and attitudes for IMU like devices (e.g. SBG Ellipse).

### DIFF
--- a/hooks/proprietary/PASHR.js
+++ b/hooks/proprietary/PASHR.js
@@ -1,0 +1,133 @@
+'use strict'
+
+/**
+ * Copyright 2016 Signal K and Fabian Tollenaar <fabian@signalk.org>.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const debug = require('debug')('signalk-parser-nmea0183/PASHR');
+const utils = require('@signalk/nmea0183-utilities');
+const moment = require('moment-timezone');
+
+/*
+=== PASHR - RT300 proprietary roll and pitch sentence ===
+Attitude/State provided by an IMU sensor.
+------------------------------------------------------------------------------
+         0           1   2    3      4      5     6     7     8   9 10 11
+         |           |   |    |      |      |     |     |     |   | |  |
+$PASHR,hhmmss.sss,hhh.hh,T,rrr.rr,ppp.pp,xxx.xx,a.aaa,b.bbb,c.ccc,d,e*hh<CR><LF>
+------------------------------------------------------------------------------
+Field Number:
+0. hhmmss.sss - UTC time
+1. hhh.hh - Heading in degrees
+2. T - flag to indicate that the Heading is True Heading (i.e. to True North)
+3. rrr.rr - Roll Angle in degrees
+4. ppp.pp - Pitch Angle in degrees
+5. xxx.xx - Heave
+6. a.aaa - Roll Angle Accuracy Estimate (Stdev) in degrees
+7. b.bbb - Pitch Angle Accuracy Estimate (Stdev) in degrees
+8. c.ccc - Heading Angle Accuracy Estimate (Stdev) in degrees
+9. d - Aiding Status
+10. e - IMU Status
+11. hh - Checksum
+*/
+
+function isEmpty(mixed){
+  return ((typeof mixed !== 'string' && typeof mixed !== 'number') || (typeof mixed === 'string' && mixed.trim() === ''))
+}
+
+module.exports = function(input){
+  const { id, sentence, parts, tags } = input;
+
+  const empty = parts.reduce((e, val) => {
+    if( isEmpty(val) ){
+      ++e
+    }
+    return e
+  }, 0);
+
+  if( empty > 3 ){
+    return null
+  }
+
+  const time = parts[0].indexOf('.') === -1 ? parts[0] : parts[0].split('.')[0];
+  const timestamp = utils.timestamp(time, moment.tz('UTC').format('DDMMYY'));
+
+  const delta = {
+    updates: [
+      {
+        source: tags.source,
+        timestamp: timestamp,
+        values: [
+          {
+            path: 'navigation.attitude',
+            value: {
+              heading: Number(parts[1]),
+              roll: Number(parts[3]),
+              pitch: Number(parts[4])
+            }
+          },
+          {
+            path: 'navigation.attitude.accuracy',
+            value: {
+              heading: Number(parts[8]),
+              roll: Number(parts[6]),
+              pitch: Number(parts[7])
+            }
+          },
+          {
+            path: 'navigation.attitude.heading',
+            value: {
+              heading: Number(parts[1]),
+              true: parts[2] === "T"
+            }
+          },
+          {
+            path: 'navigation.attitude.heading.true',
+            value: parts[2] === "T"
+          },
+          {
+            path: 'navigation.attitude.heave',
+            value: Number(parts[5])
+          },
+          {
+            path: 'navigation.attitude.aid',
+            value: utils.int(parts[9])
+          },
+
+          {
+            path: 'navigation.attitude.status',
+            value: utils.int(parts[10])
+          },
+        ]
+      }
+    ],
+  };
+
+  const toRemove = [];
+
+  delta.updates[0].values.forEach((update, index) => {
+    if( typeof update.value === 'undefined' || update.value === null || (typeof update.value === 'string' && update.value.trim() === '') || (typeof update.value === 'number' && isNaN(update.value)) ){
+      toRemove.push(index)
+    }
+  });
+
+  if( toRemove.length > 0 ){
+    toRemove.forEach(index => {
+      delta.updates[0].values.splice(index, 1)
+    })
+  }
+
+  return delta
+};
--- a/hooks/proprietary/PASHR.js
+++ b/hooks/proprietary/PASHR.js
@@ -64,6 +64,17 @@ module.exports = function(input){
   const time = parts[0].indexOf('.') === -1 ? parts[0] : parts[0].split('.')[0];
   const timestamp = utils.timestamp(time, moment.tz('UTC').format('DDMMYY'));
 
+  const GPSQualityFlags = [
+    'no GPS',
+    'All non-RTK fixed integer positions',
+    'RTK fixed integer position'
+  ];
+
+  const INSStatusFlag = [
+    'Pre-Alignment',
+    'Post-Alignment'
+  ];
+
   const delta = {
     updates: [
       {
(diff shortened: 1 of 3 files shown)


yyvus:
The PASHR grammar has been taken from the documentation attached
[NMEA0183_PASHR.pdf](https://github.com/SignalK/signalk-parser-nmea0183/files/3361984/NMEA0183_PASHR.pdf)

tkurki:
Sorry about being unresponsive, summer is for sailing.

Did you notice that the test is failing? See link to Travis run in the pull request.

yyvus:
Hi tkurki,

Yes I did see that. But I struggle to understand what is going on. I believe I have been able to set the correct parsing behaviour, but I don't grab the whole software yet and it seems to be related to the upper layers. So if you could quickly have a look it would be helpful.
This is the error I get when I run `mocha`:
```
 AssertionError: [object Object]/vessels/urn:mrn:imo:mmsi:230099999:Missing required property: url
/vessels/urn:mrn:imo:mmsi:230099999:Missing required property: uuid
:Missing required property: aircraft
:Missing required property: aton
:Missing required property: sar
/vessels/urn:mrn:imo:mmsi:230099999/navigation/attitude/accuracy:Unknown property (not in schema)
/vessels/urn:mrn:imo:mmsi:230099999/navigation/attitude/heading:Unknown property (not in schema)
/vessels/urn:mrn:imo:mmsi:230099999/navigation/attitude/heave:Unknown property (not in schema)
/vessels/urn:mrn:imo:mmsi:230099999/navigation/attitude/aid:Unknown property (not in schema)
/vessels/urn:mrn:imo:mmsi:230099999/navigation/attitude/status:Unknown property (not in schema)
/vessels/urn:mrn:imo:mmsi:230099999/navigation/attitude/value/heading:Unknown property (not in schema)
```

It obviously shows that some path are not recognized in the schema. Those paths I made up. I don't know what protocol should be followed here and I made up those paths in order to be as clear as possible... And where can we modify the schema? All this is unclear to me...

joabakk:
`toFull(delta).should.be.validSignalK` compares the paths to https://github.com/SignalK/specification

You should really PR there to get the paths aligned

tkurki:
Closing due to inactivity. Feel free to reopen if you want to push this forward.

yyvus:
Hi again,

I need to revive this NMEA 0183 PASHR sentence. So I am back working on this.

I have further identified where the problem is comming from. As suggested by @joabakk , the toFull version computed on top of the delta is wrong. I have therefore compared this full version with other message. It seems the only difference I have with the sentence GGA is here the following:
Result of `toFull` for the GGA sentence is 
'''
  vessels: {
    'urn:mrn:imo:mmsi:230099999': { mmsi: '230099999', navigation: [Object] }
  },
  self: 'urn:mrn:imo:mmsi:230099999',
  version: '0.1.0',
  sources: {
    DUMMY_LABEL: { label: 'DUMMY_LABEL', type: 'NMEA0183', GP: [Object] }
  }
}
'''
Result of `toFull` for PASHR is
'''
{
  vessels: {
    'urn:mrn:imo:mmsi:230099999': { mmsi: '230099999', navigation: [Object] }
  },
  self: 'urn:mrn:imo:mmsi:230099999',
  version: '0.1.0',
  sources: {
    DUMMY_LABEL: { label: 'DUMMY_LABEL', type: 'NMEA0183', PA: [Object] }
  }
}
'''

which differ only by the dummy label last entry. 

This last entry seems to be the `talker`, and that eventually the talker `PA` is not yet available. Do you knowwhere I should make it available?

tkurki:
Are you talking about the failing test? The problem there is not with the talker but as previously said:

>It obviously shows that some path are not recognized in the schema

yyvus:
Please help me with the code, because I don't understand your answer.

The conflicting line in passing the `npm run test` is the following: 
```
    toFull(delta).should.be.validSignalK
```
If I remove it the test passes. If I then do `console.log(toFull(delta))`, I get this:
```
{
  vessels: {
    'urn:mrn:imo:mmsi:230099999': { mmsi: '230099999', navigation: [Object] }
  },
  self: 'urn:mrn:imo:mmsi:230099999',
  version: '0.1.0',
  sources: {
    DUMMY_LABEL: { label: 'DUMMY_LABEL', type: 'NMEA0183', PA: [Object] }
  }
}
```
I understand here that the above json 'should.be.validSignalK'. Therefore, I compared it with other sentences which led to the conclusion I mentioned above. Does that make sense? If not, please explain further what is happening.

joabakk:
It will not pass before the new paths are added to signalk/specification. You should start there, and then once the specification =schema is changed, the test will pass

yyvus:
I tried to understand the schemas in the specification project. I found so far no place where any of the sentences is specified explicitly. If you could point to this place, I would be happy to change it.

yyvus:
I think I start to understand what you mean. The path I define in the PASHR.js files must appear in one of the schema specifications found in the folder schema->groups. Correct?

joabakk:
Yes correct. The failing test checks to see if the schema =specification includes the path and type. So the schema/groups needs to include the new path. I just picked a random addition to the specification. Perhaps the easiest is to start with the test file in the specification and then make the schema file. Once this test passes and the specification PR is merged, the PR test in this repo will pass...

tkurki:
I think a better course of action is to modify this so that the paths are aligned with the current spec, but not try to add them to the spec right away. And just disable the `should.be.validSignalK` test for now.

I'll add some detailed comments to the paths.

hooks/proprietary/PASHR.js
```diff
@@ -0,0 +1,144 @@
...
+*/
+
+function isEmpty(mixed){
+  return ((typeof mixed !== 'string' && typeof mixed !== 'number') || (typeof mixed === 'string' && mixed.trim() === ''))
+}
+
+module.exports = function(input){
+  const { id, sentence, parts, tags } = input;
+
+  const empty = parts.reduce((e, val) => {
```
tkurki, line 53: This should be outside decoding function, as it is static. 

I'd prefer a function named `emptyPartsCount` or something like that, so that it is immediately obvious what it does. `empty` as a variable name is actually misleading I think.

A comment why > 3 empty parts makes input invalid would also be good.

hooks/proprietary/PASHR.js
```diff
@@ -0,0 +1,144 @@
...
+  }, 0);
+
+  if( empty > 3 ){
+    return null
+  }
+
+  const time = parts[0].indexOf('.') === -1 ? parts[0] : parts[0].split('.')[0];
+  const timestamp = utils.timestamp(time, moment.tz('UTC').format('DDMMYY'));
+
+  const GPSQualityFlags = [
```
tkurki, line 67: Move these outside the decoding function also?

hooks/proprietary/PASHR.js
```diff
@@ -0,0 +1,144 @@
...
+          {
+            path: 'navigation.attitude.accuracy',
+            value: {
+              heading: Number(parts[8]),
+              roll: Number(parts[6]),
+              pitch: Number(parts[7])
+            }
+          },
+          {
+            path: 'navigation.attitude.heading',
```
tkurki, line 101: Is this different from all of the headings already in the spec? https://signalk.org/specification/1.5.0/doc/vesselsBranch.html#vesselsregexpnavigationheadingmagnetic

The current spec has separate paths for magnetic and true values. Is there reason to do things here differently?

hooks/proprietary/PASHR.js
```diff
@@ -0,0 +1,144 @@
...
+          {
+            path: 'navigation.attitude',
+            value: {
+              heading: Number(parts[1]),
+              roll: Number(parts[3]),
+              pitch: Number(parts[4])
+            }
+          },
+          {
+            path: 'navigation.attitude.accuracy',
```
tkurki, line 93: I would pick `.stdev` over `.accuracy`, because that is more obvious to interpret and documents what the valuje is. Or is accuracy something well established in this context?

hooks/proprietary/PASHR.js
```diff
@@ -0,0 +1,144 @@
...
+          {
+            path: 'navigation.attitude.heading.true',
+            value: parts[2] === "T"
+          },
+          {
+            path: 'navigation.attitude.heave',
+            value: Number(parts[5])
+          },
+          {
+            path: 'navigation.attitude.aid',
```
tkurki, line 116: What does `aid` / _aiding_ mean in this context?

hooks/proprietary/PASHR.js
```diff
@@ -0,0 +1,144 @@
...
+          },
+        ]
+      }
+    ],
+  };
+
+  const toRemove = [];
+
+  delta.updates[0].values.forEach((update, index) => {
+    if( typeof update.value === 'undefined' || update.value === null || (typeof update.value === 'string' && update.value.trim() === '') || (typeof update.value === 'number' && isNaN(update.value)) ){
```
tkurki, line 132: I'd prefer a function that is named so that it descibes the check - this is string of comparisons is hard to decipher. What is it checking?

Imho creating first an array, then a second array of things to remove and then removing them on enyb one is pretty awkward. How about `arr = arr.filter(condition)`?

tkurki (commented):
Something to consider:  if I understand this correctly the sentence contains a single attitude value with three different components and then metadata about those values. If this is correct what if we created a single `navigation.attitude` value and all the different values would be properties of the value?

hooks/proprietary/PASHR.js
yyvus: Here aid means `aiding status`. More precisely:
```
 Position mode:                    
 0 – No GPS position fix 
1 – All GPS position fixes apart from “2” 
2 – RTK Integer position fix 
```

hooks/proprietary/PASHR.js
yyvus: Aggreed. More verbose the better for me too.

hooks/proprietary/PASHR.js
yyvus: Here it is the `True Heading of the navigation system, from 0 to 359.99 degrees, using 2 decimal places`.

yyvus:
Thanks to have looked at this issue. 

Here is the detailed description of the nmea data:
pashrDetails

You can find the details here : https://www.datrontechnology.co.uk/wp-content/uploads/2016/10/nmeaman.pdf

I have spent some more time and I also had the sensation that the `navigation.attitude` should be made of all the values related to them. At least it would fit better with the specifications.

I will have more time soon to put into it.

yyvus:
What about such a structure:
```
const delta = {
    updates: [
      {
        source: tags.source,
        timestamp: timestamp,
        values: [
          {
            path: 'navigation.attitude',
            value: {
              heading: Number(parts[1]),
              roll: Number(parts[3]),
              pitch: Number(parts[4]),
              headingTrue: Boolean(parts[2] === "T"),
              heave: Number(parts[5]),
              aid: GPSQualityFlags[utils.int(parts[9])],
              status: INSStatusFlag[utils.int(parts[10])]
            }
          },
          {
            path: 'navigation.attitude.stdev',
            value: {
              heading: Number(parts[8]),
              roll: Number(parts[6]),
              pitch: Number(parts[7])
            }
          }
        ]
      }
    ],
  };
```

tkurki:
Aren't the stdev values part of the same "update"? Imho this would be more appropriate:

```
const delta = {
    updates: [
      {
        source: tags.source,
        timestamp: timestamp,
        values: [
          {
            path: 'navigation.attitude',
            value: {
              heading: Number(parts[1]),
              roll: Number(parts[3]),
              pitch: Number(parts[4]),
              stdev: {
                heading: Number(parts[8]),
                roll: Number(parts[6]),
                pitch: Number(parts[7])  
              },
              headingTrue: Boolean(parts[2] === "T"),
              heave: Number(parts[5]),
              aid: GPSQualityFlags[utils.int(parts[9])],
              status: INSStatusFlag[utils.int(parts[10])]
            }
          }
        ]
      }
    ],
  };
```

We have `navigation.headingTrue` and `navigation.headingMagnetic` but `attitude` value has `yaw`, but we have failed to specify what that value means in the specification. Nevertheless: in other parts of the data model there is an explicit path for true and magnetic value, not a value-plus-qualifier structure.

yyvus:
I don't know if a new `path` is necessary for the `stdev`. I don't understand the difference between a `path` and a `value` yet. So I follow your suggestion on this matter.

tkurki:
Closing as stale. Feel free to reopen if needed.

PR #155, opened 2019-08-19
SeaTalk composite data management

raffmont:
SeaTalk produces different datagrams for position (0x50, 0x51) and for datetime (0x54, 0x56).
This pull request fixes the issue https://github.com/SignalK/signalk-parser-nmea0183/issues/154 about the navigation.position and (topic not covered by the issue, but related to the same behaviour) navigation.datatime.
In order to solve the composite data management, two module global data structure are used:
position - for latitude and longitude
datetime - for date (year, month, day) and time (hour, minute, seconds, milliSeconds)
In 0x50 and 0x51 hooks, a SignalK update is created if and only if both latitude and longitude have been already read.
In 0x54 and 0x56 hooks, a SignalK update is created when both data and time have been already read.

### DIFF
--- a/hooks/ALK.js
+++ b/hooks/ALK.js
@@ -30,11 +30,11 @@ STALK     	Raymarine Seatalk1 datagram sentence
 
 const seatalkHooks = require('./seatalk')
 
-module.exports = function(input) {
+module.exports = function(input, session) {
   const { id, sentence, parts, tags } = input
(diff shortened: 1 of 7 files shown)


tkurki:
Module globals will not work correctly if the system has more than one STALK source. See https://github.com/SignalK/signalk-parser-nmea0183/blob/master/hooks/VDM.js#L51-L53 on how to use a `session` for managing state.

PR #156, opened 2019-08-19
SeaTalk composite data: implementation with session

raffmont:
The use of 0x50, 0x51 (latitude, longitude) and 0x54, 0x56 (time, date) datagrams have been implemented using the session propagated by the Parser class.

### DIFF
--- a/hooks/ALK.js
+++ b/hooks/ALK.js
@@ -30,11 +30,11 @@ STALK     	Raymarine Seatalk1 datagram sentence
 
 const seatalkHooks = require('./seatalk')
 
-module.exports = function(input) {
+module.exports = function(input, session) {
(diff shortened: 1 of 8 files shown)


tkurki:
STALK tests are now failing, you probably got a message about them? Please fix: make sure `npm test` succeeds on your machine.

raffmont:
The tests failed because not compliant with the new 0x50 0x51 behaviour and because 0x54 and 0x56 have been disabled in order to prevent a misbehaviour.
I updated the tests in order to match the new implementation.

tkurki:
No need to fix the commits, easy enough to use Github SquashAndMerge, since this can be just a single commit.

PR #164, opened 2020-02-18
  delete signs before $ if $ is availible 

Thomas-GeDaD:
some devices like NASA Clipper Wind sends an /x00 before the sentence. Because of that I add the $ control at sentence start if it is available

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -55,8 +55,7 @@ class Parser {
       sentence = sentence.substr(sentence.indexOf("$") , sentence.length)
     }
     let valid = utils.valid(sentence, this.options.validateChecksum)
-    
-    
+       
     if (valid === false) {
       throw new Error(`Sentence "${sentence.trim()}" is invalid`)
     }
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -49,17 +49,21 @@ class Parser {
     if (typeof tags.timestamp === 'undefined') {
       tags.timestamp = new Date().toISOString()
     }
-
-    let valid = utils.valid(sentence, this.options.validateChecksum)
     
+    //delete signs before $ if $ is availible
+    if (sentence.indexOf("$") !== -1) {
```
sbender9: You should store the result of `indexOf` away here so that you don't do it twice:

```
let idx = sentence.indexOf("$")
if ( idx !== -1 ) {
  ...
}
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@signalk/nmea0183-signalk",
-  "version": "3.6.0",
+  "version": "3.6.1",
```
sbender9: Please don't update the version number since you don't know which version this would go into. That will be done when we publish a new version.

lib/index.js
sbender9: You should also check for zero, no need to do the substring if at zero.

```
if ( idx !== -1 && idx !== 0 ) {
...
}
```

Thomas-GeDaD:
O.K. I change this next time. But at the Moment i cannot test becouse havn´t a device now.

Issue #166, opened 2020-02-22
VDM Meassage 21 AtoN request; Off position and Virtual.


Hakansv:
From VDM message 21 AtoN the "off-position indicator" and "Virtual AtoN flag" are missing.
It's sufficient to flag for either "off position" or "virtual". Absence of either of are understood as "on position" or "Real" respectively. 

Reference: [Message 21 example](https://www.navcen.uscg.gov/?pageName=AISMessage21)
![bild](https://user-images.githubusercontent.com/7202854/75097402-b3595980-55aa-11ea-9ce5-be56e366a589.png)
![bild](https://user-images.githubusercontent.com/7202854/75097387-8a38c900-55aa-11ea-9718-5da24a5ad624.png)
![bild](https://user-images.githubusercontent.com/7202854/75097395-a2104d00-55aa-11ea-82cb-dc5c66231eb5.png)

OCPN method to parse VDM message 21:
```
int offpos = bstr->GetInt( 260, 1 ); // off position flag
int virt = bstr->GetInt( 270, 1 ); // virtual flag
```

sbender9:
I've added off position indicator: #168 

But the decoder we use does not support the virtual flag it would need to be added there first:

https://github.com/fulup-bzh/GeoGate/tree/master/encoder

Hakansv:
@sbender9 
The "off position" flag is very good and appreciated. Thanks.

The virtual flag is possible to, although uncertain, detect from the mmsi number. (xxyyy6zzz) The "6" here is a recommended to use for a Virtual AtoN. But since it's a recommendation to use for states involved it can't be all time verified. A secure flag would be the one from the AtoN itself.
If it's possible to request that support from your decoder team we would be more safe even with data flowing though SK? 

There are so far two possible type of AtoN supported in message 21. "Real, "Virtual" ->270, 0/1. 
(There is a (future?) third type "Mobile" mentioned but to support that in message 21 it has to be redefined I think.)

afischerdev:
I'll add the virtual flag in the encoder lib.
Do you have a test sentence with a positive flag?

Hakansv:
@afischerdev 
Not so I exactly know where to find it for the moment. But I'll find, just give me a day or two and I be back.
Thanks
Håkan

afischerdev:
@Hakansv 
Thanks, but not requiered anymore, I found one in the current test data.

Hakansv:
[...]

PR #169, opened 2020-02-24
feature: add support for aton virtual flag

sbender9:
See #166

### DIFF
--- a/hooks/VDM.js
+++ b/hooks/VDM.js
@@ -224,6 +224,12 @@ module.exports = function (input, session) {
         value: { "id": data.aidtype, "name": atonType }
       })
     }
+    if ( typeof data.virtual !== 'undefined' ) {
+      values.push({
+        path: 'virtual',
+        value: data.virtual == 1
+      })
+    }
   }
 
   if ( data.cargo )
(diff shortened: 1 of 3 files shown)


sbender9:
https://github.com/SignalK/specification/issues/563

tkurki:
Rebased on master, but some tests fail.

sbender9:
It says all checks passed??

tkurki:
Still fails locally for me.

tkurki:
Just a local glitch.

Hakansv:
@sbender9  @tkurki 
I think the logic for the virtual flag has got the opposite value.
This one is Real.
Håkan
image

sbender9:
Do you have the sentence for that one?

Hakansv:
Here is one of each
https://github.com/SignalK/signalk-parser-nmea0183/issues/166#issuecomment-592054206

Hakansv:
@sbender9 
Here goes dragons...
For the two examples in the file above the virtual flags seems correct. (They are from Belgium)
My comment was about two Swedish AtoN that has got opposite flag from what have been observed before.
My earlier indication is taken from the mmsi number saying a "6" in position like: xxzzz6xxx but the own indication from the VDM message would be more reliable.
A further investigation has shown you data is correct. The SE authorities has probably misused the mmsi number. 
And when I compare to the use of VDM data in OCPN where we have the same message parsing as you have used all is correct.
Please close this comment. Sorry for the trouble.
Thanks
Håkan

PR #174, opened 2020-04-27
add $ recognition

Thomas-GeDaD:
add $ recognition, delete signs before $ if $ available and not at the beginning of the sentence
checked with nasa Clipper instruments

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,9 +49,14 @@ class Parser {
     if (typeof tags.timestamp === 'undefined') {
       tags.timestamp = new Date().toISOString()
     }
-
-    let valid = utils.valid(sentence, this.options.validateChecksum)
     
+    //delete signs before $ if $ is availible and not on start pos.
+    let idx = sentence.indexOf("$")
+    if ( idx !== -1 && idx !== 0 ) {
+      sentence = sentence.substr(idx , sentence.length)
+    }
+    let valid = utils.valid(sentence, this.options.validateChecksum)
+       
     if (valid === false) {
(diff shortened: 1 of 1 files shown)


Thomas-GeDaD:
Hi, have edit the recognition as described. Have tested with Nasa Clipper NMEA0183 Instruments and several further NMEA0183 devices.

joabakk:
What is the purpose of this change, and what does it solve? Please explain

Thomas-GeDaD:
some NMEA devices like Nasa Clipper sends a 00 character before the $ sign at the beginning. SignalK can not parse this correctly. 
See this issue:
https://github.com/SignalK/signalk-server-node/issues/970

tkurki:
I have another, more generic solution planned for this. Stay tuned.

tkurki:
I would rather not add an extra conditional for every single user of this library just because some equipment misbehaves.

I added a PR for an option to explicitly remove NULL characters at https://github.com/SignalK/signalk-server-node/pull/1032. This will cause no overhead if the option is not turned on.

Thomas-GeDaD:
solved with 👍 
https://github.com/SignalK/signalk-server-node/pull/1032
Hi tkurki, thanks for your support. What speaks against it to enable this option permanently in the code? I be afraid of that users irritated through this option?

tkurki:
Just what I had said above:

>I would rather not add an extra conditional for every single user 

I would guess 99.9% of devices don't produce nulls.

The server's processing pipeline consists of several steps, sort of like unix command line pipes consist of several components. I would rather clean up input before it gets to the parser.

PR #179, opened 2020-10-11
feature: add MTA,  MWD  and VWT sentences

pbfulmar:
Hi,
integrated a NKE instrument system with SignalK. 
Added the following sentences (all sent by the NKE Multigraphic) to the NMEA parser:
- MTA, mean temperature of air
- MWD, wind speed and direction
- VWT, true wind angle and speed

Thanks for all your great work with SignalK!

### DIFF
--- a/hooks/MTA.js
+++ b/hooks/MTA.js
@@ -42,7 +42,7 @@ module.exports = function (input) {
         timestamp: tags.timestamp,
         values: [
           {
-            path: 'environment.air.temperature',
+            path: 'environment.outside.temperature',
             value: utils.transform(utils.float(parts[0]), 'c', 'k')
             //returns raw value, no transformation done
           }
--- a/hooks/MTA.js
+++ b/hooks/MTA.js
@@ -1,0 +1,55 @@
+/**
+ * Copyright 2016 Signal K <info@signalk.org> and contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+ const utils = require('@signalk/nmea0183-utilities')
+
+ /*
+  * MTA - Mean Temperature of Air
+  *
+  *        0   1 2
+  *        |   | |
+  * $--MTA,x.x,C*hh<CR><LF>
+  *
+  * Field Number:
+  *   0.    Degrees
+  *   1.    Unit of Measurement, Celcius
+  *   2.    Checksum
+  *
+  */
+
+module.exports = function (input) {
+  const { id, sentence, parts, tags } = input
+
+  const delta = {
+    updates: [
+      {
+        source: tags.source,
+        timestamp: tags.timestamp,
+        values: [
+          {
+            path: 'environment.air.temperature',
+            value: utils.transform(utils.float(parts[0]), 'c', 'k')
+            //returns raw value, no transformation done
+          }
+        ]
+      }
+    ],
+  }
+
+  return delta
+}
--- a/hooks/MTA.js
+++ b/hooks/MTA.js
@@ -33,8 +33,11 @@
   */
 
 module.exports = function (input) {
-  const { id, sentence, parts, tags } = input
+  const { id, sentence, parts, tags } = input;
 
+  if(parts[1] != 'C') {
+      return null;
+  }
   const delta = {
     updates: [
       {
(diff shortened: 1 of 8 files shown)


hooks/MTA.js
```diff
@@ -0,0 +1,55 @@
...
+'use strict'
+
+ const utils = require('@signalk/nmea0183-utilities')
+
+ /*
+  * MTA - Mean Temperature of Air
+  *
+  *        0   1 2
+  *        |   | |
+  * $--MTA,x.x,C*hh<CR><LF>
```
tkurki, line 26: Since the unit field is in the sentence I think we should add a check that the value is actually `C` and not something else.

test/MWD.js
```diff
@@ -0,0 +1,36 @@
...
+
+chai.use(require('chai-things'))
+
+describe('MWD', () => {
+
+  it('speed & direction data', () => {
+      const delta = new Parser().parse('$IIMWD,,,046.,M,10.1,N,05.2,M*0B')
+
+    delta.updates[0].values.should.contain.an.item.with.property('path', 'environment.wind.speedTrue')
+    delta.updates[0].values[1].value.should.be.closeTo(5.19585, 0.00005)
```
tkurki, line 31: The test data contains 5.2 meters per second, so why do we convert from knots? Which one is more correct?

test/MWD.js
```diff
@@ -0,0 +1,36 @@
...
+const should = chai.Should()
+
+chai.use(require('chai-things'))
+
+describe('MWD', () => {
+
+  it('speed & direction data', () => {
+      const delta = new Parser().parse('$IIMWD,,,046.,M,10.1,N,05.2,M*0B')
+
+    delta.updates[0].values.should.contain.an.item.with.property('path', 'environment.wind.speedTrue')
```
tkurki, line 30: I know other tests are written like this, but I think it is bad practices, as this test doesn't catch the case where you have accidentally messed up the order of the paths. If we are testing the values by position in the values array I think we can also test paths like that.

Would you mind redoing the path tests so that they are like the value test?

test/VWT.js
```diff
@@ -0,0 +1,25 @@
...
+const chai = require('chai')
+const should = chai.Should()
+
+chai.use(require('chai-things'))
+
+describe('VWT', () => {
+  it('True wind converts ok', () => {
+    const delta = new Parser().parse('$IIVWT,030.,R,10.1,N,05.2,M,018.7,K*75')
+//'$IIMWV,074,T,05.85,N,A*2E'
+// $IIVWT,030.,R,10.1,N,05.2,M,018.7,K*75
```
tkurki, line 13: Why the commented lines?

hooks/MWD.js
```diff
@@ -0,0 +1,71 @@
...
+ * <5> N = Knots
+ * <6> Wind speed, meters/second, to the nearest 0.1 m/s.
+ * <7> M = Meters/second
+ */
+
+module.exports = function(input) {
+  const { id, sentence, parts, tags } = input
+  var pathValues = []
+
+  if(parts[0] != ''){
```
tkurki, line 41: Since the sentence structure seems to provide for having the fields in different order I'd like to include also tests for the fields 1 and 3.

tkurki (changes requested):
Thanks for the pr - looks pretty good, some minor review comments.

hooks/MTA.js
pbfulmar: Done. Also adding similar checks to the other two sentences (MWD, VWT)

test/MWD.js
pbfulmar: Not at all, agree. Done.

test/VWT.js
pbfulmar: only overlooked to cleanup :-) 
deleted.

test/MWD.js
pbfulmar: Got a point there. I changed the implementation to give preference to the m/s value and changed the test so equality is used when possible and closeTo when the value had to be converted.

hooks/MWD.js
pbfulmar: Ok, I added the tests for all the fields. 
Are you sure the fields are allowed in different order?  From the definition I would assume the fields (data/type) for True/Magnetic/Knots/Meters per Second must be in exactly this sequence.

pbfulmar:
Thank you for your kind reply and the detailed and helpful review! 
I simply started with copy-paste of some existing modules since I couldn't tell which checks are eventually already done somewhere else, maybe by the parser.

The new commits include: 
- checks for the units/designator fields in the MTA, MWD and VWT sentences
- converted values (from knots or km/h) are only used, when no value in m/s is present in any of the three sentences
- tests for the pathes are changed to positional references and a set of more systematic tests

This is my very first pr in github :-) should I click on the "resolve conversation" buttons or are you going to do that? 
Viele Grüße
Peter

tkurki:
And a fine one it is!

I have absolutely no idea if the fields can be in different order, but when not knowing exactly maybe the extra tests are in order.

If you want to tweak this to perfection in the ideal world this would be three separate commits, that each add one sentence, but I can merge it like it is and just squash all the commits to one. I don't want to see the whole story, just the end  results in master's commit history.

pbfulmar:
Thank you! 
And I will remember to keep any future pr as small as possible :-)

Regarding the order of the fields: at least for MWD it must be fixed because 'M' is used for both 'magnetic' as well as for 'm/s'.  In different order it would be ambiguous. Better don't trust any input data - so I think the extra tests are definitely in order.

Anything else you'd like me to do?

PR #188, opened 2021-01-25
Add PSMDST 

joabakk:
Same as $STALK, but official proprietary sentence for seatalk.


test/PSMDST.js
```diff
@@ -0,0 +1,214 @@
...
+  })
+
+  it('0x50 Latitude converted', () => {
+    const delta = new Parser().parse(latitude)
+    delta.updates[0].values.should.contain.an.item.with.property('path', 'navigation.position.latitude')
+    delta.updates[0].values[0].value.should.be.closeTo(33, 0.5)
+  })
+
+  it('0x51 Longitude converted', () => {
+    const delta = new Parser().parse(longitude)
```
tkurki, line 115: Oh, this is a problem with the existing Seatalk support. I assume all clients work with the path in spec: `navigation.position` and an object value. This will not be treated as a position, with two separate messages with a non spec path.

Do you know how the data stream works in the real world? Could the parser keep the latest value of one or the other and trigger emit on the other, making sure that the value of the other field is not too old?

hooks/proprietary/PSMDST.js
```diff
@@ -0,0 +1,41 @@
...
+1 			hex       	First datagram content
+2 			hex   		Last datagram content
+3 			hex      	Checksum
+*/
+
+const seatalkHooks = require('../seatalk')
+
+module.exports = function(input) {
+  const { id, sentence, parts, tags } = input
+  const key = '0x' + (parts[0]).toUpperCase()
```
tkurki, line 35: Is `toUpperCase` really needed?

joabakk: I don't think so, but if some DIY seatalk talker would emit a 0x9c datagram, it would fail. To my knowledge, official seatalk is always uppercase

test/PSMDST.js
joabakk: This is a copy of the ALK.js code. The actual code seems to be ok from https://github.com/SignalK/nmea0183-signalk/pull/156, but the tests may not have followed?

tkurki: tests/ALK.js different https://github.com/SignalK/nmea0183-signalk/blob/master/test/ALK.js#L110-L121

That works, but the position test is not the greatest, because it relies on the tests running in order - they are not independent. All of this is actually just one test https://github.com/SignalK/nmea0183-signalk/blob/master/test/ALK.js#L108-L121 and should be structured as one.

If the tests for ALK and PSMDST are the same, except for the inputs, it would be great to restructure them so that there is only one copy of the tests, running for two sets of test data.

I tried to fix the Gitub test action to run on PRs, could you please rebase on master so that we can see that tests run.

joabakk:
Messed up repo, closing and re-opening #189

PR #189, opened 2021-02-02
Add $PSMDST

joabakk:
Same as $STALK, replaces #188


joabakk:
I am struggling with the "sessions". The position and time tests pass for $STALK, but for some reason not for $SPMDST

tkurki:
Simple: ALK.js uses session, PSMDST.js doesn't.

joabakk:
> Simple: ALK.js uses session, PSMDST.js doesn't.

Argh. Thanks again

PR #190, opened 2021-02-11
Add custom sentences for motor diagnostic

norbert-walter:
I add custom sentences for motor diagnostic. See the project at:
https://open-boat-projects.org/en/diy-motordiagnose/


norbert-walter:
Hello, I see the test unit are not valid. How I can start the test on my local system? I need to program any test units.

joabakk:
I don't think it is a good idea to mask custom sentences as proprietary sentences. 

There is a long list of vendor codes in https://www.nmea.org/Assets/20130801%200183%20identifier%20list.pdf. Even if there is no conflict between these sentences and valid official proprietary sentences, I think we should avoid this. 

Many of us have started with nmea0183 as a way to transport information, but this is better covered by Signal K as a common denominator.

joabakk:
> Hello, I see the test unit are not valid. How I can start the test on my local system? I need to program any test units.

If you 'npn install' in your local repo folder, you can 'npm test'

norbert-walter:
Do you want to accept the changes? Otherwise I won't make the effort to make changes.

joabakk:
I would wait for the discussion to pan out before making any changes. I am only airing my initial thoughts. Perhaps if you explain the use case in more detail and why this is the best solution, we can agree on a path forward

norbert-walter:
For NMEA0183 there are the user-defined Sentences. In SignalK, this is also intended to include such Sentences. Some are available for NKE devices. Why shouldn't one also allow others? I have built a device for engine diagnosis. It should also work natively on old NMEA0183 buses. Not all information can then be used, but the most important information is available. With the user-defined Sentences I expand the possibilities to read in additional data. Legacy and current third-party software also support custom tenants. NMEA0183 over TCP or UDP is the most common simple interface in third party software. The possibility of importing user-defined Sentences in SignalK enables e.g. data transfer to the NMEA2000 bus. At the boat exhibition "Boot 2020" in Düsseldorf at Open Boat Projects, many sailors have problems exchanging data between the two bus worlds. The nets are very heterogeneous in old boats. Only data feed via SignalK reduces the usability outside of SignalK.

hooks/proprietary/PECWT.js
```diff
@@ -0,0 +1,73 @@
...
+            path: 'propulsion.engine_' + parts[1] + '.CoolingWaterTemp',
+            value: {
+              description: 'Cooling water temp in [K]',
+              units: 'K'
+             }
+          }
+        ],
+        values: [
+          {
+            path: 'propulsion.engine_' + parts[1] + '.CoolingWaterTemp',
```
sbender9, line 64: This should be coolantTemperature and is a know key so does not need meta.

hooks/RPM.js
```diff
@@ -30,20 +30,32 @@
...
+            path: `propulsion.engine_${parts[1]}.pitch`,
+            value: {
+              description: `Propeller pitch in [%]`,
+              units: `%`
+             }
+          }
+        ],
         values: [
           {
             path: `propulsion.${(parts[0].toUpperCase() === 'S' ? 'shaft' : 'engine')}_${parts[1]}.revolutions`,
```
sbender9, line 40: There is an existing place in the spec for this, see https://github.com/SignalK/specification/blob/master/schemas/groups/propulsion.json

hooks/proprietary/PEERT.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+  const delta = {
+    updates: [
+      {
+        source: tags.source,
+        timestamp: tags.timestamp,
+        meta: [
+          {
+            path: 'propulsion.engine_' + parts[1] + '.EngineRoomTemp',
```
sbender9, line 55: This wold normally be at environment/inside/engineRoom/temperature

hooks/proprietary/PEELT.js
```diff
@@ -0,0 +1,68 @@
...
+            path: 'propulsion.engine_' + parts[1] + '.EngineLifeTime',
+            value: {
+              description: 'Engine live time in [h]',
+              units: 'h'
+             }
+          }
+        ],
+        values: [
+          {
+            path: 'propulsion.engine_' + parts[1] + '.EngineLifeTime',
```
sbender9, line 59: Please use camelCase here and elsewhere.

sbender9 (changes requested):

sbender9:
Note that I don't really have an opinion on weather we should things this way. @tkurki what do you think?

sbender9:
Maybe a way to have custom sentences defined in separate packages and loaded dynamically?

joabakk:
> For NMEA0183 there are the user-defined Sentences. In SignalK, this is also intended to include such Sentences. Some are available for NKE devices. Why shouldn't one also allow others? I have built a device for engine diagnosis. It should also work natively on old NMEA0183 buses. Not all information can then be used, but the most important information is available. With the user-defined Sentences I expand the possibilities to read in additional data. Legacy and current third-party software also support custom Senteces. NMEA0183 over TCP or UDP is the most common simple interface in third party software. The possibility of importing user-defined Sentences in SignalK enables e.g. data transfer to the NMEA2000 bus. At the boat exhibition "Boot 2020" in Düsseldorf at Open Boat Projects, many sailors have problems exchanging data between the two bus worlds. The nets are very heterogeneous in old boats. Only data feed via SignalK reduces the usability outside of SignalK.

There are no user-defined sentences in NMEA0183, only proprietary sentences, meant for manufacturers to add their measurements that do not fit. The NKE sentence is an example of a documented proprietary sentence. My issue is that if we start adding private workarounds to a parser meant for a specific protocol, we are peeing in our pants to keep warm. 

If you are sending custom sentences over the 0183 bus, you need to control the software both in the sender and the receiver. That means you are only using the cable or network link. This is where I started out as well. 

Of course you should get the data in Signal K, but that can be achieved by other means than this hack.

My opinion is that we should not mix custom and proprietary sentences, and if we were to do it, we would have to take an unused mnemonic code such as 'SIK', be consistent (all $PSIK..) and hope that there is no conflict. 

Annoying as it may be to have to argue for a suggestion or work in different ways, this is one of the many strengths of open source, the end result will be better. I'm still open to be persuaded.

hooks/proprietary/PEELT.js
norbert-walter: What is camelCase? I don't know.

norbert-walter:
If you look at a NMEA0183 bus system, there are only a few participants in the bus (<10). The user himself has control over which hardware he installs and uses. The NMEA0182 to SignalK gateway knows the standard services. Usually only a few proprietary devices are used, the exact functionality of which is known. Conflicts with other devices will not occur, especially since SignalK knows exactly what the Sentences are and can handle them correctly. With wind sensors and speed sensors, sometimes crude extensions are made and you have to make wild contortions in the software for decoding. The shortcoming of XDR Sentences is that the content is difficult to decode if it is complex, as the data structure can change dynamically. I've experimented with XDR myself and I'm not happy with it. Hence the special custom Sentences. They are unambiguous, have a fixed structure and can be easily decoded. Custom Sentences should actually only be reloaded when necessary. Unfortunately, there is no easy way to do this in SignalK. Custom Sentences are currently implemented. They can be implemented permanently if they can be activated later via the configuration. Maybe that is an option? This avoids conflicts even with multiple occupancy of the same or similar Sentences. In practice, one does not use several Custom Sentences at the same time. I think it would be good if you could add and activate Custom Sentences. I will add an alternative way of directly feeding SignalK into my device for those who primarily use SignalK.

hooks/proprietary/PEERT.js
norbert-walter: Ok, I change to the other path.

hooks/RPM.js
norbert-walter: I add only the information for pitch. The pitch is not defined and need the meta information to unit. In most cases we dont use the pitch. My last contact with a motorboat driver needs the pitch information. He has two bigger engines and want to control both maschines.

hooks/proprietary/PECWT.js
norbert-walter: Ok I change to path coolantTemperature.

hooks/proprietary/PEELT.js
sbender9: The first character should should be lowerCase.

hooks/RPM.js
sbender9: Pitch is there. propulsion.*.drive.propeller.pitch

norbert-walter: Ok, I change to this path

hooks/proprietary/PEELT.js
norbert-walter: I use lower characters.

sbender9: `EngineLifeTime` should be `engineLifeTime`

norbert-walter: Ah ok, sorry

norbert-walter:
A new version is transmitted

norbert-walter:
Is it now ok with the sentences and Paths for SignalK? Then I make the test units. Wich files I need to create for the test units?

tkurki:
I have an idea for a mechanism how one could add custom sentences to the nmea0183 parser from a plugin. I'll get back to this over the weekend.

tkurki:
Started, but not there yet. Will get back to this in a few days.

norbert-walter:
Ok, thanks. I waiting.

norbert-walter:
Any ideas or comments?

norbert-walter:
Is something else coming?

tkurki:
Slowly, but yes.

tkurki:
Finally got back to this. #193 has the mechanism (requires matching changes in the server).

tkurki:
@norbert-walter can you take a look at #193? It needs the server's master version, as the changes there are not published yet.

tkurki:
No replies and there's now a custom parser mechanism in place. Let me know if you need help with that.

norbert-walter:
I have lost the desire to contribute my ideas.

tkurki:
My sincere apologies if that is due to my actions / inactions. Please reconsider.

tkurki:
I totally understand that it feels frustrating. I was not involved in the beginning of this PR and can only account for my own actions.

Again, all I can offer is my apology for time wasted and hope that we can do better in the future.

PR #203, opened 2021-09-06
Add Parser ts types

MrBlenny:
I've added some types for the `Parser` class. I noticed you're using typescript in some other SignalK projects so didn't think it'd be too controversial to add these types directly in this repo. I'm happy to add them to DefinitelyTyped if you prefer.

One thing to note, I couldn't quite figure out what was goin on with `emitPropertyValue` so I didn't include that in the constructor interface.

Also, note
```ts
export interface IParserDelta {
  updates: {
    source: string;
    timestamp: string;
    values: {
      path: string;
      value: any;
    }[];
  }[];
}
```
Includes `source` and `timestamp`. These are not included in what is returned from `custom-sentence-plugin`. Not sure if this should be optional or whether `custom-sentence-plugin` should be changed.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -52,7 +52,7 @@ export interface IParserOptions {
   onPropertyValues?: (
     propertyName: string,
     onPropertyValuesCallback: IParserOptionsOnPropertyValuesCallback,
-  ) => any;
+  ) => void;
   validateChecksum?: boolean;
 }
 
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,64 @@
+export interface IParserDelta {
+  updates: {
+    source: string;
+    timestamp: string;
+    values: {
+      path: string;
+      value: any;
+    }[];
+  }[];
+}
+
+export interface IParserDeltaWithSource {
+  updates: {
+    source: {
+      sentence: string;
+      talker: string;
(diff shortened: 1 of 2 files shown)


tkurki:
PropertyValues related types are already ts in https://github.com/SignalK/signalk-server/blob/master/packages/server-api/src/propertyvalues.ts https://www.npmjs.com/package/@signalk/server-api albeit in very "raw" format, just the code and types published without any accompanying documentation.

MrBlenny:
I hadn't seen signal-ts thanks!
Would you like me to change this PR to use any of these existing types?

I feel adding `signalk-server` as a dependency could lead to circular dep issues and `signal-ts` includes `joi` for runtime validation which isn't required here. That said, it could make sense if the `IParserDelta` interface was imported from `SignalK/specification`? Could be interesting to see what running [json-schema-to-typescript](https://www.npmjs.com/package/json-schema-to-typescript) produces.

tkurki:
https://www.npmjs.com/package/@signalk/server-api is separate from `signalk-server` and has zero dependencies, so there's no danger of circular deps.

Deltas are very simple beasts and have been pretty stable, I don't think going for json-schema-to-typescript makes sense.

tkurki:
Closing for inactivity, feel free to reopen if needed.

PR #204, opened 2021-09-21
fix: ZDA parser chokes when time has zero seconds

sarfata:
When seconds are 0 the parser would try to divide by seconds. Milliseconds are then NaN and the time is invalid.


tkurki:
@signalk/nmea0183-signalk@3.10.1

PR #205, opened 2021-09-26
Fix sign of cross track error obtained from RMB sentence

ardbiesheuvel:
XTE.js has this

  const direction = parts[3].trim().toUpperCase() === 'L' ? 1 : -1

signalk-to-nmea0183/sentences/RMB.js has this

      return nmea.toSentence([
        '$IIRMB',
        crossTrackError.toFixed(2),
        crossTrackError < 0 ? 'R' : 'L',
        ...

RMB.js does the opposite, and produces a negative value when the required
correction is to steer left. The definition in XTE.js appears to produce the
expected result in combination with my autopiot and the NMEA2000 conversions
as well, so let's change RMB.js to match XTE.js

Signed-off-by: Ard Biesheuvel <ardb@kernel.org>

### DIFF
--- a/hooks/RMB.js
+++ b/hooks/RMB.js
@@ -68,7 +68,7 @@ module.exports = function (input) {
   crossTrackError = utils.float(parts[1])
   crossTrackError = (!isNaN(crossTrackError)) ? crossTrackError : 0.0
 
-  crossTrackError = parts[2] == 'R' ? crossTrackError : -crossTrackError;
+  crossTrackError = parts[2] == 'L' ? crossTrackError : -crossTrackError;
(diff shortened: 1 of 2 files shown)


tkurki (changes requested):
Nice!

You need to change the checksum in the test input sentences.

ardbiesheuvel:
> Nice!
> 
> You need to change the checksum in the test input sentences.

Ah yes, of course. Will fix and resubmit

tkurki (approved):

tkurki:
Published in 3.10.2. Thank you!

PR #214, opened 2021-10-30
RMC now returns 'null' when field value is missing as per specification

davidsanner:
Where fields were missing values RMC.js returned 0 incorrectly instead of null as per this in the specification

Under: Missing or invalid data
https://github.com/SignalK/specification/blob/c446914b331084dc5c8bcddf6aa65003b69b1c45/gitbook-docs/data_model.md#missing-or-invalid-data

eg. variation should not be set to 0.0 if RMC is field is empty.  Same goes for lat/lon and others.

### DIFF
--- a/hooks/RMC.js
+++ b/hooks/RMC.js
@@ -36,75 +36,68 @@ values:
  -      *6A          The checksum data, always begins with *
 */
 
-module.exports = function (parser, input) {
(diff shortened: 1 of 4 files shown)


davidsanner:
Closing and submitting PR with current version

PR #216, opened 2021-10-30
RMC will now output null when missing/invalid values not 0

davidsanner:
When 0183 fields were missing values RMC.js returned 0 incorrectly instead of null as per the specification

Under: Missing or invalid data
https://github.com/SignalK/specification/blob/c446914b331084dc5c8bcddf6aa65003b69b1c45/gitbook-docs/data_model.md#missing-or-invalid-data

eg. variation should not be set to 0.0 if RMC is field is empty. Same goes for lat/lon and others.

(3rd time the charm?  other PR were stuck on old version)

### DIFF
--- a/hooks/RMC.js
+++ b/hooks/RMC.js
@@ -39,25 +39,23 @@ values:
 module.exports = function (input) {
   const { id, sentence, parts, tags } = input
 
-  let latitude = -1
-  let longitude = -1
-  let speed = 0.0
-  let track = 0.0
-  let variation = 0.0
+  let latitude = null
+  let longitude = null
+  let speed = null
+  let track = null
+  let variation = null
 
   const timestamp = utils.timestamp(parts[0], parts[8])
   const age = moment.tz(timestamp, 'UTC').unix()
 
-  latitude = utils.coordinate(parts[2], parts[3])
-  longitude = utils.coordinate(parts[4], parts[5])
+  latitude = !isNaN(parseFloat(parts[2])) && isFinite(parts[2]) && "NS".includes(parts[3]) ? utils.coordinate(parts[2], parts[3]) : null
+  longitude =!isNaN(parseFloat(parts[4])) && isFinite(parts[4]) && "EW".includes(parts[5]) ? utils.coordinate(parts[4], parts[5]) : null
 
-  speed = utils.float(parts[6])
-  speed = !isNaN(speed) && speed > 0 ? speed : 0.0
+  speed = !isNaN(parseFloat(parts[6])) && isFinite(parts[6]) && parts[6] >= 0 ? utils.transform(parts[6], 'knots', 'ms') : null
 
-  track = utils.float(parts[7])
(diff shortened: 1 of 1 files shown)


hooks/RMC.js
```diff
@@ -39,25 +39,23 @@ values:
...
+  let speed = null
+  let track = null
+  let variation = null
 
   const timestamp = utils.timestamp(parts[0], parts[8])
   const age = moment.tz(timestamp, 'UTC').unix()
 
-  latitude = utils.coordinate(parts[2], parts[3])
-  longitude = utils.coordinate(parts[4], parts[5])
+  latitude = !isNaN(parseFloat(parts[2])) && isFinite(parts[2]) && "NS".includes(parts[3]) ? utils.coordinate(parts[2], parts[3]) : null
```
tkurki, line 51: `isFinite(NaN)` is false, so `isNan` is not really needed and adds an extra parseFloat to everything?

tkurki (changes requested):
Instead of changing the existing test data to match the assertion we should change the assert AND add a new test case for the value being zero, please.

hooks/RMC.js
davidsanner: Agreed, just wasn't familiar with the proper/preferred way to test for null with Chai.    Latest commit has two test sentences which check for valid & null entries with SOC/COG & variation.

tkurki: I think you read my comment out of context? I meant that we don't need both `!isNaN` and `isFinite` in the ternary's condition.

davidsanner: Since javascript coerces empty strings to 0 and !isNaN('') returns true it wasn't straight forward with a single test.  My original though for testing was:   
parseFloat(parts[2])).trim().length > 0 && !isNaN(parseFloat(parts[2]))
I looked around for the best/preferred way to determine if a string was a valid number but I see what I found has issues.
I can update with the test above or if you have a preferred way please let me know.  Thanks.

davidsanner: Err:  parts[2]).trim().length > 0 && !isNaN(parts[2])
Seems so simple  ;)

davidsanner: I went ahead and committed the data validity check shown above.  Hopefully that looks good.  If not happy to replace with a preferred method & wrap up this small but important change for RMC.

tkurki:
@signalk/nmea0183-signalk@3.10.5

Hakansv:
@tkurki 
Would it be possible to update SK with this PR?
Please compare: https://www.cruisersforum.com/forums/f134/weird-stuff-happening-opencpn-5-5-1109-on-openplotter-and-buster-rpi-3-b-257870.html#post3528026

tkurki:
@Hakansv This is merged and published as version 3.10.5 of the module @signalk/nmea0183-signalk. But people won't get the latest version, untill we publish a new version of the server core that will prompt people to update the server - this module will get also updated as part of a server update.

PR #219, opened 2022-02-01
change four values from int to float to match GGA sentence

davidsanner:
When horizontalDilution was < 1.0 it was being set to zero causing troubles.  3 other values (MSL Altitude ,Geoid Separation, & Age of differential GPS data) were being set to int when the GGA sentence accepts float.


tkurki:
Published as @signalk/nmea0183-signalk@3.10.6

PR #220, opened 2022-02-02
Rounding new GGA floats to 1 decimal place as SK/JS stores up to 17

davidsanner:
Seeing a few 17 place decimals in GGA, rounding to single decimal as shown in sentence description
eg. SK has:  navigation.gnss.horizontalDilution   0.7000000000000001

### DIFF
--- a/hooks/GGA.js
+++ b/hooks/GGA.js
@@ -121,17 +121,17 @@ module.exports = function (input) {
 
           {
             path: 'navigation.gnss.antennaAltitude',
-            value: utils.float(parts[8]),
+            value: Math.round(utils.float(parts[8]) * 10) / 10 ,
           },
 
           {
             path: 'navigation.gnss.horizontalDilution',
-            value: utils.float(parts[7]),
+            value: Math.round(utils.float(parts[7]) * 10) / 10 ,
           },
 
           {
             path: 'navigation.gnss.geoidalSeparation',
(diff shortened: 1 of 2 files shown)


davidsanner:
This may be a larger issue than just this plugin as values from N2K sources also have a huge number of "non-significant" digits.  This will help if sentence isn't suppressed from 0183 events (goes direct to 10110 output) but I think an update signalk-to-nmea0183 is needed as well to make sure the GGA sentences  (and others?) are well formed.  Multiple values with 17 decimal places could break things.

joabakk:
I would suggest only fixing it where it matters, ie in the clients or the 0183 output.

davidsanner:
Many legacy 0183 clients/equipment can't be updated and I would think 17 decimal digits might break some of them and agree it may make more sense to format data on the output. I've done that with this (mentioned above) PR https://github.com/SignalK/signalk-to-nmea0183/pull/85  but what about source events that pass directly through and don't use the signalk-to-nmea0183 plugin?   Small issue and hopefully someone has a clear idea of the fix but we don't want to break GGA.  If there's a better solution / direction I can close this PR.

tkurki:
I think we can leave conversion to Signal K as is and just make sure that conversions to other formats have some sane amount of decimal places. Your fix in signalk-to-nmea0183 is now published btw.

PR #223, opened 2022-04-15
feature: do not fill timestamp

### LINKED ISSUE #222: NMEA sentences with an included timestamp set the server timestamp value
When receiving NMEA records that include a timestamp this value is used to set the "timestamp" field of the data in signalk. Records that do not include a timestamp use the servers local time.
This way it is nearly impossible to determine the age of a signalK data value as you don't know if the time reference is the gps time or the time of the signalK server (as they easily could differ).
Remark:
NMEA2000 data seems to always use the local server timestamp.

tkurki:
The contract of nmea0183-signalk should be to convert
source data to Signal K and only source data. If it
fills in timestamp downstream processing can not distinguish
between timestamp from source data and timestamp filled
in on the fly with current time.

Fixes #222.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -68,10 +68,6 @@ class Parser {
       tags = {}
     }
 
-    if (typeof tags.timestamp === 'undefined') {
-      tags.timestamp = new Date().toISOString()
-    }
-
     let valid = utils.valid(sentence, this.options.validateChecksum)
 
     if (valid === false) {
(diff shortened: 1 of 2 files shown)


joabakk:
This change may make downstream plugins have to change their logic, but that's fair. Makes sense

tkurki:
How? Server fills in timestamp for deltas that do not have one.

joabakk:
Ok my bad. Then all should be fine

PR #225, opened 2022-07-31
Added support for RPM and Engine Runtime from Cruzpro RH30/60/110 - Productcode B

SamHoulder:
http://www.cruzpro.com/rh60.html
Decoding from https://www.electronicspoint.com/forums/threads/engine-hour-meter-with-nmea-output.159207/


tkurki:
Thank you!

PR #228, opened 2022-09-21
Initial work on XDR parser

fabdrol:
This PR implements an initial implementation for an XDR parser that uses the definitions format from `xdr-parser-plugin`. The implementation is different however, as that plugin has several issues IMO (e.g. `require` statements within function bodies). 

In order to finish this PR, it would make sense (to me) to internalise the dictionary of XDR sentences rather tha pulling it from `xdr-parser-plugin`, and expanding on the definitions in there. Also, it would make sense to make the expression dependent on the unit (RN I'm simply checking if the units match).

### DIFF
--- a/hooks/XDR.js
+++ b/hooks/XDR.js
@@ -1,0 +1,13 @@
+// $IIXDR,C,C,10.7,C,AIRTEMP,A,0.5,D,HEEL,A,-1.-3,D,TRIM,P,1.026,B,BARO,A,A,-4.-3,D,RUDDER*18
+
+const xdrDictionary = require('xdr-parser-plugin/xdrDict');
+const xdrParser = require('xdr-parser-plugin/xdrParser');
+
+
+module.exports = function (input) {
+  const { sentence } = input
+
+  const delta = xdrParser(sentence, xdrDictionary);
+  console.log(`Parsing ${sentence} >>> ` + JSON.stringify({ delta }, null, 2));
+  return delta
+}
--- a/hooks/XDR.js
+++ b/hooks/XDR.js
@@ -1,7 +1,7 @@
 // $IIXDR,C,C,10.7,C,AIRTEMP,A,0.5,D,HEEL,A,-1.-3,D,TRIM,P,1.026,B,BARO,A,A,-4.-3,D,RUDDER*18
 
+const math = require('math-expression-evaluator');
 const fs = require('fs');
-const xdrParser = require('xdr-parser-plugin/xdrParser');
 const xdrDictionary = { definitions: [] };
 
 // Populate a dictionary
@@ -21,7 +21,7 @@ if (fs.existsSync(xdrDictPath)) {
 xdrDictionary.definitions = [
   ...xdrDictionary.definitions,
   {
-    type: "Air temperature",
+    type: "value",
     data: "temperature",
     units: "C",
     name: "AIRTEMP",
@@ -29,17 +29,17 @@ xdrDictionary.definitions = [
     sk_path: "environment.outside.temperature",
   },
   {
-    type: "Roll",
-    data: "Angles",
-    units: "Deg",
+    type: "roll",
+    data: "angle",
+    units: "D",
     name: "HEEL",
     expression: "(x*pi/180)",
     sk_path: "navigation.attitude"
   },
   {
-    type: "Outside temperature",
-    data: "Angles",
-    units: "Deg",
+    type: "value",
+    data: "angle",
+    units: "D",
     name: "RUDDER",
     expression: "(x*pi/180)",
     sk_path: "steering.rudderAngle"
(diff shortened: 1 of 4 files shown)


tkurki:
Where would the user configure this? The nice thing about plugins is that configuration ui is essentially free.

Also are you aware of https://github.com/SignalK/nmea0183-signalk#custom-sentences ?

fabdrol:
No, wasn't ware. But, this is a PR to expand the functionality of this parser, not just for users of Signal K server.

hooks/XDR.js
```diff
@@ -0,0 +1,160 @@
...
+
+    if (value === null || typeFlag === null || unit === null) {
+      continue
+    }
+
+    if (isNaN(parseFloat(value))) {
+      continue
+    }
+
+    const def = definitions.find(({ name }) => (name === boundary));
```
tkurki, line 114: If there is ever going to be only one active definition per name, so why not make `definitions` an object? Then this would be `definitions[name]`.

The current structure doesn't support having multiple formulas for multiple units. Another object layer keyed by unit would allow `definitions[name][unit]` lookup, supporting multiple units per definition.

hooks/XDR.js
```diff
@@ -0,0 +1,160 @@
...
+    const prevBoundary = index === 0 ? null : boundaries[index - 1];
+    const elements = [];
+    let fill = false;
+
+    for (p of parts.slice(1)) {
+      if (!fill && (!prevBoundary || p === prevBoundary) && elements.length === 0) {
+        fill = true;
+      }
+
+      if (fill === false || p === boundary) {
```
tkurki, line 80: `!fill`

hooks/XDR.js
```diff
@@ -0,0 +1,160 @@
+// $IIXDR,C,C,10.7,C,AIRTEMP,A,0.5,D,HEEL,A,-1.-3,D,TRIM,P,1.026,B,BARO,A,A,-4.-3,D,RUDDER*18
+
+const math = require('math-expression-evaluator');
+const fs = require('fs');
+const xdrDictionary = { definitions: [] };
```
tkurki, line 5: What's the idea with having `definitions` and not just definitions directly under root?

hooks/XDR.js
```diff
@@ -0,0 +1,160 @@
...
+      continue
+    }
+
+    const def = definitions.find(({ name }) => (name === boundary));
+
+    if (!def) {
+      continue;
+    }
+
+    if (def.units !== unit) {
```
tkurki, line 120: I think both should be in singular or plural.

hooks/XDR.js
```diff
@@ -0,0 +1,160 @@
+// $IIXDR,C,C,10.7,C,AIRTEMP,A,0.5,D,HEEL,A,-1.-3,D,TRIM,P,1.026,B,BARO,A,A,-4.-3,D,RUDDER*18
+
+const math = require('math-expression-evaluator');
+const fs = require('fs');
+const xdrDictionary = { definitions: [] };
+
+try {
+  // Populate a dictionary
+  const xdrDictPath = require.resolve('xdr-parser-plugin/xdrDict');
```
tkurki, line 9: I am not terribly fond of having a configuration file that is loaded using `require.resolve`. I think configuration should be injected from the context that uses the parser. We can support both though if you think this is the way to go, just that we should not log a warning if there is no configuration file.

test/XDR.js
```diff
@@ -0,0 +1,27 @@
...
+      'environment.outside.temperature'
+    )
+    delta.updates[0].values[0].value.should.be.closeTo(283.85, 0.005)
+
+    delta.updates[0].values.should.contain.an.item.with.property(
+      'path',
+      'navigation.attitude.roll'
+    )
+    delta.updates[0].values[1].value.should.be.closeTo(0.0087, 0.00005)
+  })
```
tkurki, line 26: How about some non happy path tests - a malformed expression, input with no matching definition?

hooks/XDR.js
```diff
@@ -0,0 +1,160 @@
...
+    type: "value",
+    data: "angle",
+    units: "D",
+    name: "RUDDER",
+    expression: "(x*pi/180)",
+    sk_path: "steering.rudderAngle"
+  }
+]
+
+module.exports = function (input) {
```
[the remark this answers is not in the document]
fabdrol: I'll look into this, good suggestion

hooks/XDR.js
fabdrol: This behaviour is mostly a result of me attempting to re-use the definitions from `xdr-parser-plugin`, which became less-and-less tenable as I went through development. All good points.

hooks/XDR.js
fabdrol: See previous comment regarding `xdr-parser-plugin`

hooks/XDR.js
fabdrol: See previous comment regarding `xdr-parser-plugin`

tkurki:
Closings as stale.

PR #232, opened 2022-11-22
feature: add GNS sentence

RoeeDev:
Hi,
I Added the GNS sentence to the NMEA parser, hope I did it well.
Thanks for all your great work with SignalK!

### DIFF
--- a/hooks/GNS.js
+++ b/hooks/GNS.js
@@ -1,0 +1,193 @@
+'use strict'
+
+/**
+ * Copyright 2016 Signal K and Fabian Tollenaar <fabian@signalk.org>.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const debug = require('debug')('signalk-parser-nmea0183/GGA')
+const utils = require('@signalk/nmea0183-utilities')
+const moment = require('moment-timezone')
+
+/*
+=== GNS - Fix Data ===
+Time, Position and fix related data for a GPS receiver.
+------------------------------------------------------------------------------
+        0         1       2 3        4 5    6  7   8   9   10  11  12
+        |         |       | |        | |    |  |   |   |   |   |   |
+ $--GNS,hhmmss.ss,ddmm.mm,a,dddmm.mm,a,c--c,xx,x.x,x.x,x.x,x.x,x.x*hh<CR><LF>
+------------------------------------------------------------------------------
+Field Number:
+0. UTC of position, hh is hours, mm is minutes, ss.ss is seconds
+1. Latitude, dd is minutes, mm.mm is minutes
+2. N or S (North or South)
+3. Longitude, dd is minutes, mm.mm is minutes
+4. E or W (East or West)
+5. Mode indicator (non-null) - Variable character field with one character for each supported constellation:
+    * First character is for GPS
+    * Second character is for GLONASS
+    * Third character is Galileo
+    * Fourth character is for BeiDou
+    * Fifth character is for QZSS
+    * Subsequent characters will be added for new constellations
+   Each character will be one of the following
+     - N = No fix. Satellite system not used in position fix, or fix not valid
+     - A = Autonomous. Satellite system used in non-differential mode in position fix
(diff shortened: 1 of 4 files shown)


hooks/GNS.js
```diff
@@ -0,0 +1,193 @@
...
+
+function isEmpty(mixed) {
+  return (
+    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
+    (typeof mixed === 'string' && mixed.trim() === '')
+  )
+}
+
+function indicator(chars, modes) {
+    const systems = ["GPS","GNSS","Galileo","BeiDou","QZSS"]
```
tkurki, line 76: GNSS => Global Navigation Satellite System, this should be GLONASS here.

But I think decoding the pretty compact and machine readable string to a single human oriented string is a disservice. We should have something like 
```
{ 
   GPS: 'Precise',
   Galileo: 'Autonomous'
}
```
- only include the ones that are operational - we don't want to list all of them in each GNS message, if the poor GNSS device has only GPS for example
- define short but human readable enumerated strings for the modes
- use an object structure to capture this data instead of a string
-

tkurki (changes requested):
Nice work! The only problem is that some of this stuff is not covered by the Signal K specification, which is probably the reason this was never added here.

See my more detailed comments.

hooks/GNS.js
```diff
@@ -0,0 +1,193 @@
...
+            path: 'navigation.gnss.differentialAge',
+            value: utils.float(parts[10]),
+          },
+
+          {
+            path: 'navigation.gnss.differentialReference',
+            value: Number(parts[11]),
+          },
+          {
+            path: 'navigation.state',
```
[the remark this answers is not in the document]
RoeeDev: Do by isValidSignalKVessek you mean the following function:
[https://github.com/SignalK/nmea0183-signalk/blob/6e2602c0570c7121f493a1c6ab9b29f8aa2b29c1/lib/index.js#L127-L133](url)

Because I couldn't find a function with this specific name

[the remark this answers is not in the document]
tkurki: To be exact: you should see `toFull(delta).should.be.validSignalK` failing when you output paths that are not in the Signal K specification.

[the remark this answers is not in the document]
RoeeDev: Ok yes I see it happen. So should I just remove this line from the GNS file located in the test directory? Or is there another place to change in order to disable the validation?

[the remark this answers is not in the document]
tkurki: Comment it out please

hooks/GNS.js
```diff
@@ -0,0 +1,194 @@
...
+*/
+
+function isEmpty(mixed) {
+  return (
+    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
+    (typeof mixed === 'string' && mixed.trim() === '')
+  )
+}
+
+function indicator(chars, modes) {
```
tkurki, line 75: Modes are always the same, right? If they never change no need to pass them as a parameter, let's make a constant with name all in CAPITAL letters. Same for systems. And place them nicely right before `indicator`.

hooks/GNS.js
```diff
@@ -0,0 +1,194 @@
...
+    "E": "Estimated",
+    "F": "RTK Float",
+    "M": "Manual",
+    "N": "No Valid Fix",
+    "P": "Precise",
+    "R": "RTK Integer",
+    "S": "Simulator"
+  }
+
+  const status = {
```
tkurki, line 111: Constant name should be all caps.

hooks/GNS.js
```diff
@@ -0,0 +1,194 @@
...
+    (typeof mixed === 'string' && mixed.trim() === '')
+  )
+}
+
+function indicator(chars, modes) {
+    const systems = ["GPS","GLONASS","Galileo","BeiDou","QZSS"]
+    const indications = {}
+    chars.forEach(function (val, i) { indications[systems[i]] = modes[val] })
+    return indications
+  }
```
tkurki, line 80: you could write this in a more idiomatic form like

```
return chars.reduce( (acc, c, i) => {
  acc[SYSTEMS[i]] = MODES[c]
  return acc
}, {})
```

tkurki (changes requested):
Looking good - a few more stylistic fixes, if you will. Sorry about the piecemeal review.

hooks/GNS.js
RoeeDev: It's all good - I'm happy to learn best practices :)

tkurki:
👍 - now just fix it so it/tests work and we're done!

RoeeDev:
Fixed the test, I had a small typo in the test file.
Hopefully it should work now.

tkurki:
You can not compare objects in the test with `value.should.equal`, it should be `value.should.deep.equal`. https://stackoverflow.com/questions/13225274/the-difference-between-assert-equal-and-assert-deepequal-in-javascript-testing-w

The checksum for the empty sentence is not correct, must end with `*32`. https://nmeachecksum.eqth.net/

tkurki (approved):

PR #233, opened 2022-12-02
Added changes to accomodate ShipModul Miniplex

jprotzman:
ShipModul now uses C and R in their PSMDST message to represent a message being sent or received.


jprotzman:
I realized I should add code to handle both the new format as well as the old. Closing this PR until that code is complete.

PR #234, opened 2022-12-02
Added support for ShipModul Miniplex Latest Firmware

jprotzman:
The ShipModul Miniplex has added a field to the PSMDST message to indicate sending to 'C' or receiving from 'R' the SeaTalk network. This change adds support for the new field by looking for the 'R' field then parsing the data accordingly.

### DIFF
--- a/hooks/proprietary/PSMDST.js
+++ b/hooks/proprietary/PSMDST.js
@@ -22,7 +22,7 @@
 $PSMDST,Z,xx,yy,nn*CS
 where:
 PSMDST     	Raymarine Seatalk1 datagram sentence
-0       C/R       R for Recevied messages, C for sent messages
+0       C/R       R for Recevied messages, C for sent messages *Note: This field only exists in later firmware versions of the ShipModul Miniplex
 1 			00-9C     Datagram type
 2 			hex       First datagram content
 3 			hex   		Last datagram content
@@ -33,8 +33,10 @@ const seatalkHooks = require('../seatalk')
 
 module.exports = function (input, session) {
   const { id, sentence, parts, tags } = input
-  const key = '0x' + parts[1].toUpperCase();
-  input.parts = parts.slice(1, input.parts.length);
+  if (parts[0].toUpperCase() === 'R') {
+    input.parts = parts.slice(1, input.parts.length);
+  }
+  const key = '0x' + input.parts[0].toUpperCase();
   if (typeof seatalkHooks[key] === 'function') {
     return seatalkHooks[key](input, session)
   } else {
--- a/hooks/proprietary/PSMDST.js
+++ b/hooks/proprietary/PSMDST.js
@@ -22,7 +22,7 @@
 $PSMDST,Z,xx,yy,nn*CS
 where:
 0       PSMDST     	Raymarine Seatalk1 datagram sentence
-1       C/R       R for Recevied messages, C for sent messages *Note: This field only exists in later firmware versions of the ShipModul Miniplex
+1       C/R       R for Received messages, C for sent messages *Note: This field only exists in later firmware versions of the ShipModul Miniplex
 2 			00-9C     Datagram type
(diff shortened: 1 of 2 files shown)


test/seatalk.js
```diff
@@ -53,7 +53,7 @@ const should = chai.Should()
 chai.use(require('chai-things'))
 
 describe('seatalk', () => {
-  ;['$PSMDST,', '$STALK,'].forEach((prefix) => {
+  ;['$PSMDST', '$PSMDST_R', '$STALK,'].forEach((prefix) => {
```
tkurki, line 56: If I understood this PR correctly this test is not really relevant: it now tests the normal input format, but for `PSMDST_R`. Shouldn't we actually add a test where input does indeed have the extra `R` in the data?

Does the parser ever see the `C` data?

tkurki:
Is this PR still alive?

test/seatalk.js
jprotzman: I thought this test was testing all of the typical input formats. It previously had both $PSMDST and $STALK which are typical from various devices. The new Miniplex firmware adds the _R for all received data so this test just adds that option as well.

Also, the _C field may appear at the parser but I cannot imagine these would ever need to be parsed since they are sent messages meant to go onto the ST network.

jprotzman:
Apologies for the slow response.

test/seatalk.js
tkurki: Several different things:
- You added support to a new data format, where the first (optional) data field has value`R`. That looks like it works, but none of the tests exercise it - you are running the tests that were already in place, with exact the same data fields as previously, no `R` field, just for an additional prefix
- the tests do not pass for `PSMDST` prefix because you deleted the trailing comma in the prefix value (see the diff above, it is highlighted)
- the tests do not pass for `PSMDST_R` prefix because there is no trailing comma in the new prefix value

hooks/proprietary/PSMDST.js
```diff
@@ -17,22 +17,26 @@
...
+0  1  2  3  4
+|  |  |  |  |
+$PSMDST,Z,xx,yy,nn*CS
 where:
 PSMDST     	Raymarine Seatalk1 datagram sentence
-0 			00-9C       	Datagram type
-1 			hex       	First datagram content
-2 			hex   		Last datagram content
-3 			hex      	Checksum
+0       C/R       R for Recevied messages, C for sent messages *Note: This field only exists in later firmware versions of the ShipModul Miniplex
```
tkurki, line 25: typo in _Recevied_.

Could you please line the field number markers up with the fields like so
image

hooks/proprietary/PSMDST.js
```diff
@@ -17,22 +17,26 @@
...
+3 			hex   		Last datagram content
+4 			hex      	Checksum
 */
 
 const seatalkHooks = require('../seatalk')
 
 module.exports = function (input, session) {
   const { id, sentence, parts, tags } = input
-  const key = '0x' + parts[0].toUpperCase()
+  if (parts[0].toUpperCase() === 'R') {
```
tkurki, line 36: I think we should structure the code so that
- if there is no extra field parse as before this PR - test on the number of fields
- if there is the extra field
  - if it is something else return undefined so that no output is generated
  - if it is 'R' remove it and parse as before this PR

tkurki (changes requested):

tkurki:
Still there?

jprotzman:
Yes still here. Just haven't had time to review yet.

test/seatalk.js
jprotzman: This should be corrected now.

tkurki (approved):

tkurki:
If you have any more details than "latest firmware", like the exact version when the R was added and links to Miniplex documentation feel free to add to the PR description or as a comment here. This PR will be linked from the release notes, so somebody might even check details here.

PR #237, opened 2022-12-17
fix: HDG deviation handling

### LINKED ISSUE #235: HDG sentence incorrectly parsed
As mentioned in https://github.com/SignalK/specification/issues/242#issuecomment-1345666329 , the HDG sentence is incorrectly parsed:

If I send the following sentence `$INHDG,180,5,W,10,W*6D`

navigation.magneticDeviation is not set (should be -5)
navigation.magneticVariation is set to -10 (correct)
navigation.headingCompass is not set (should be 180)
navigation.headingMagnetic is set to 180 (should be 175)
navigation.headingTrue is not set (should be 165)

Incidentally, the [NMEA spec](https://www.plaisance-pratique.com/IMG/pdf/NMEA0183-2.pdf) clearly states that the first value in the HDG sentence is the magnetic sensor reading **before** any deviation correction is applied, and **not** the magnetic heading.

tkurki:
Fixes #235.

Previously deviation was not handled at all. Now if there is deviation it is applied to create nav.headingMagnetic and in addition nav.headingCompass is output. This PR also adds deviation to the output.

If there is no deviation value the output is as previously: the sensor value in the first field is output as is as nav.headingMagnetic and no nav.headingCompass. This is a bit backwards, as what is being output is nav.headingCompass, but I think changing that would warrant a major version change for being so incompatible with the current behavior.

If the sentence includes variation also nav.headingTrue is output.

### DIFF
--- a/hooks/HDG.js
+++ b/hooks/HDG.js
@@ -39,21 +39,48 @@ function isEmpty(mixed) {
 }
 
 module.exports = function (input) {
-  const { id, sentence, parts, tags } = input
-
+  const { parts, tags } = input
   const values = []
-  if (!isEmpty(parts[0])) {
+
+  const headingCompass = parts[0]
+  const deviation = parts[1]
+  const deviationDir = parts[2] === 'E' ? 1 : -1
+  const variation = parts[3]
+  const variationDir = parts[4] === 'E' ? 1 : -1
+  if (!isEmpty(headingCompass)) {
+    const effectiveDeviation = !isEmpty(deviation) ? Number(deviation) * deviationDir : 0
     values.push({
       path: 'navigation.headingMagnetic',
-      value: utils.transform(utils.float(parts[0]), 'deg', 'rad'),
+      value: utils.transform(utils.float(headingCompass) + effectiveDeviation, 'deg', 'rad'),
     })
+    if (!isEmpty(deviation)) {
+      values.push({
+        path: 'navigation.headingCompass',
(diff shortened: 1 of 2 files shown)


tvr256:
I almost agree with the proposed logic, except for when there is no deviation value. In that case would it make sense to set both headingCompass and headingMagnetic to the same value?

tvr256:
Also, we should be able to populate nav.headingTrue by adding Variation to headingMagnetic

tkurki:
Per my slack comment: as I understand it nowadays very few devices output deviation and variation data in HDG sentence and the data is already corrected per the sensors internal calibration. Adding the exact same value as a duplicate with another path for this, the most common case, would provide very little value, only extra data traffic.

I was going to say that derived data plugin provides the magneticTrue calculation and people can do that, but in the case where variation is in the HDG data it makes perfect sense to do it here (added). Again I expect that in the real world this rarely happens, but mostly variation comes from a GPS RMC sentence and derived data plugin is the solution that will work there.

tvr256:
Understood, and your approach sounds like a sensible compromise.

If a compass doesn't output deviation it will either leave the deviation field empty or set it to 0. Should we treat 0 deviation the same as empty deviation, and only populate nav.headingCompass if it's different to nav.headingMagnetic?

tkurki:
In theory a compass could be outputting deviation 0 for some headings. If deviation is there we should probably use it at face value and not try to be too smart about it.

Issue #253, opened 2024-04-16
Relative humidity value is % rather than ratio.


panaaj:
The value assigned to Relative humidity is a percentage value (e.g 80) rather than ratio (i.e. 0.8) as per the spec.

PR #254, opened 2024-04-17
Fix: convert salinity and relativeHumidity  in VDM from percentage values to ratio

panaaj:
**Issue:** (#253)
`salinity` and `relativeHumidity` paths in the specification (`environment` group) are defined with the unit `ratio`.

Values from the VDM sentence are written to these signal K paths as a number representing a percentage (e.g. 80) rather than a ratio (i.e. 0.8).

This PR converts the received values to a ratio so they align with the specification.


tkurki:
Something is off with the VDM test timestamp??

panaaj:
Fixed the test.

PR #255, opened 2024-06-07
Fix: not sending delta when next position values are missing

panaaj:
Fix RMB, BWC and BWR sentence processing behaviour to send a delta when position field values are missing.

**_Current behaviour:_**
Whe position field values are missing, function does not return data and no delta is sent.
This means that clients are left unaware of a previously active destination being "de-activated" and will continue to show the destination as active.

**_FIx:_**
When position field values are missing, return data so that a delta is sent with `nextPoint.position` value = null.

### DIFF
--- a/test/RMB.js
+++ b/test/RMB.js
@@ -75,6 +75,9 @@ describe('RMB', () => {
 
   it("Doesn't choke on empty sentences", () => {
     const delta = new Parser().parse('$ECRMB,,,,,,,,,,,,,*77')
-    should.equal(delta, null)
+    delta.updates[0].values.should.contain.an.item.with.property(
+      'path',
+      'navigation.courseRhumbline.nextPoint.position'
+    )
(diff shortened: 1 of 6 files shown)


test/RMB.js
```diff
@@ -75,6 +75,9 @@ describe('RMB', () => {
 
   it("Doesn't choke on empty sentences", () => {
     const delta = new Parser().parse('$ECRMB,,,,,,,,,,,,,*77')
-    should.equal(delta, null)
+    delta.updates[0].values.should.contain.an.item.with.property(
+      'path',
+      'navigation.courseRhumbline.nextPoint.position'
+    )
```
tkurki, line 81: Add an assertion for the value being null?

tkurki:
Now that I think of it: at least BWC BWR also map to nextPoint, they should also be updated.

Issue #262, opened 2025-10-06
Very sporadic navigation.position.latitude = null


jaffadog:
Moving this issue here (nmea0183-signalk) from signalk-server:
https://github.com/SignalK/signalk-server/issues/2098

Some AIS/GPS units produce malformed nmea0183 messages that contain invalid latitude or longitude - yet the message has a valid checksum.

Examples from a Vespex XB-6000:

````
$GPRMC,210735.00,A,1547\x0E70800,S,14506.50460,W,0.187,10.33,110925\f12.49,E,A*3E
$GPRMC,164942.00,A,155&.61154,S,14503.46725,W,0.033,60.41,12092%,12.49,E,D*32
$GPRMC,215519.20,A,1556.60809,R,14503.46686,W,0.251,78.93,120925,12.49,D,D*3F
$GPRMC,181041.40,A,1556.61 82,S,14503.46676,W,0.152,258.79,120925,1".49,E,D*0E
````

When the nmea message is parsed and validated, the position should be further validated. If either latitude or longitude is null or out of range, the entire position should be discarded and not published in the SK data model (navigation.position).

PR #263, opened 2025-10-07
Don't publish navigation.position when lat or lng is invalid

jaffadog:
* update parsing of common GPS position NMEA sentences so that they dont publish navigation.position when either latitude or longitude is invalid (#262)
*bugfix: geoidalSeparation not being reported correctly in GGA 
*update test cases

Note this update is dependent on an updated version of https://github.com/SignalK/nmea0183-utilities. A separate PR has been filed for that package (https://github.com/SignalK/nmea0183-utilities/pull/23). I have bumped the package.json dependency version to 0.9.0 here - same as package.json of that other PR:
````
"@signalk/nmea0183-utilities": "^0.9.0"
````

### CODE REVIEW

**Comment by tkurki** [comment]
Hmm. In general in SK if a sensor is working but is not able to produce data we send out null value, that will "write over" the last value. Should we just ignore invalid position values, or send out nulls? Thoughts @panaaj @sbender9 ?

**Comment by tkurki** [comment]
`value: null` to indicate ”there is no valid value for navigation.position (but comms between you and the source for this data is working)”

**Comment by tkurki** [comment]
The former, per my message above

**Comment by jaffadog** [comment]
Humm 🤔, some of these nmea0183 parsers (like [GGA](https://github.com/SignalK/nmea0183-signalk/blob/cb59d68cbb48b0d62f7dff493e97a5cacd3d658b/hooks/GGA.js#L156), GNS, ZDA) have code like the sample below - which is designed to remove nulls from the delta. It's been this way since the initial commit. Investigating further...

````
  const toRemove = []

  delta.updates[0].values.forEach((update, index) => {
    if (
      typeof update.value === 'undefined' ||
      update.value === null ||
      (typeof update.value === 'string' && update.value.trim() === '') ||
      (typeof update.value === 'number' && isNaN(update.value))
    ) {
      toRemove.push(index)
    }
  })

  if (toRemove.length > 0) {
    toRemove.forEach((index) => {
      delta.updates[0].values.splice(index, 1)
    })
  }

  return delta
````

**Comment by jaffadog** [comment]
@tkurki - I have updated the PR such that it now publishes:

```
path: 'navigation.position',
value: null,
```

Let me know if you need any other changes.

PR #274, opened 2025-11-12
Sentence $PASHR

kiete:
Add the support of PASHR sentence by Trimble Field Systems
https://help.fieldsystems.trimble.com/r780/nmea0183-messages-pashr.htm

I haven't seen any examples of  pitch and roll implementation in Signal K so I've speculated on these keys (in Radian).

```
balance.rollAngle
balance.pitchAngle
balance.rollAngleAccuracy
balance.pitchAngleAccuracy
```

Also, I don't know how to interpret the `T` flag, so I've chosen to not set the TrueHeading if absent.

### DIFF
--- a/hooks/PASHR.js
+++ b/hooks/PASHR.js
@@ -1,0 +1,87 @@
+/**
+ * Copyright 2019 Signal K and Fabian Tollenaar <fabian@decipher.industries>.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const debug = require('debug')('signalk-parser-nmea0183/PASHR')
+const utils = require('@signalk/nmea0183-utilities')
+
+/**
+ * $PASHR,225444.123,125.12,T,12.45,11.91,xxx.xx,1.123,1.456,2.653,1,4*38
+ *
+ * Roll and pitch information - https://help.fieldsystems.trimble.com/r780/nmea0183-messages-pashr.htm
+ *
+ * 0)     225444.123        UTC time of fix 22:54:44.123
+ * 1,2)   125.12,T          True Heading, flag to indicate that the Heading is True Heading
+ * 3)     12.45             Roll Angle in degrees
+ * 4)     11.91             Pitch Angle in degrees
+ * 5)     xxx.xx            Reserved Field (Not Supported)
+ * 6)     1.123             Roll Angle accuracy estimate in degrees
+ * 7)     1.456             Pitch Angle accuracy estimate in degrees
+ * 8)     2.653             Heading Angle accuracy estimate in degrees
+ * 9)     1                 GNSS Quality [0-6]
+ * 10)    4                 IMU Alignment Status [0-4]
+ **/
+
+
+module.exports = function PASHRHook(input) {
+  const { id, sentence, parts, tags } = input
+
+  debug(`[PASHRHook] decoding sentence ${id} => ${sentence}`)
+
+  let state = {
+    timestamp: utils.timestamp(parts[0]),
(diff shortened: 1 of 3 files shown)


msallin:
Is that (at least) partially the same as: 
https://signalk.org/specification/1.5.0/doc/vesselsBranch.html#vesselsregexpnavigationattitude

Here in action (from an IMU and pypilot)
image

kiete:
> Is that (at least) partially the same as: https://signalk.org/specification/1.5.0/doc/vesselsBranch.html#vesselsregexpnavigationattitude
> 
> Here in action (from an IMU and pypilot) image

Thanks,  I will came back with the update

msallin:
Thanks for the PR, and sorry for the slow turnaround. Review follows — several items need addressing before this is merge-ready. Structured by aspect; skip anywhere you disagree and we can discuss.

### 1. Functional correctness

- **Signal K paths are non-standard.** `balance.rollAngle`, `balance.pitchAngle`, and their `*Accuracy` siblings aren't in the spec. As flagged earlier, the spec defines [`navigation.attitude`](https://signalk.org/specification/1.5.0/doc/vesselsBranch.html#vesselsregexpnavigationattitude) with `roll`, `pitch`, `yaw` (all radians). Please rename:
  - `balance.rollAngle` → `navigation.attitude.roll`
  - `balance.pitchAngle` → `navigation.attitude.pitch`
- **No standard Signal K path for roll/pitch/heading *accuracy*.** `navigation.headingTrueAccuracy`, `balance.rollAngleAccuracy`, `balance.pitchAngleAccuracy` are all invented here. I'd suggest dropping them from v1 of this sentence and opening a separate discussion on the Signal K spec repo. Inventing paths in a parser is the wrong layer — once clients depend on them they're painful to rename. If you really need them now, at minimum namespace them under `sensors.*` or a clearly vendor-specific path, not under `navigation.*`.
- **Timestamp sourced from `parts[0]` produces today's date for replayed data.** `utils.timestamp('225444.123')` returns `<today>T22:54:44.000Z` because the sentence carries only HHMMSS. Every other hook in `hooks/` uses `tags.timestamp` (set by the Parser from the tag block or the ingestion time). Please switch to `tags.timestamp` — otherwise logs replayed tomorrow will be misdated.
- **No guard against an all-empty sentence.** When every field is empty, the hook still emits a delta with six nulls. Other hooks (`HDT.js`, `HDG.js`, `HSC.js`) return `null` in that case; please do the same — emitting null deltas is noise.
- **Doc comment checksum is wrong.** The header example shows `*38`; real checksum for that sentence is `*56` (verified by XOR). Since your test uses `*56`, just fix the header to match.
- **`parts[2] == 'T'` should be `===`.** Matches the rest of the codebase and prettier/lint expectations.

### 2. Maintainability

- **Copyright header says 2019.** Looks like copy-paste from `HSC.js`. New file — the test file correctly says 2025; align `PASHR.js` to the same.
- **`README.md` not updated.** The Supported sentences list needs a PASHR entry (author's Trimble link is fine).
- **Prettier violations.** Master recently pinned `trailingComma: "none"` (dada988). Your object/array literals have trailing commas and will fail `npm run prettier:check`. Run `npm run prettier` before pushing.
- **Rebase needed.** The branch is a few commits behind master (chai 6, prettier 3, `@signalk/nmea0183-utilities` ^0.12.0). Please rebase onto `master` so CI runs clean.
- **`state` object is built only to be read once.** Minor, but you can drop the intermediate and inline, which is how the other hooks are written. Not a blocker.

### 3. Tests

- **`closeTo(2, 18166156, 0.01)` is a typo — the test is trivially passing.** Chai's `closeTo(expected, delta)` takes two args; your third arg is silently ignored. You meant `closeTo(2.18166156, 0.01)` (decimal point, not comma). Same bug on `closeTo(0, 2712, 0.01)` for `rollAngle` and `closeTo(0, 2078, 0.01)` for `pitchAngle`. As-is, any finite number within ±18 million rad/±2712 rad passes. This needs fixing — it's the whole point of the test.
- **Dead commented-out block at the top of `test/PASHR.js`.** Please remove.
- **Indentation is off** on the second and third `it()` blocks. Prettier will fix it.
- **Missing trailing newline** in `test/PASHR.js`.
- **Suggest adding** a test asserting that a fully-empty sentence returns `null` (after you add the guard above).

### Summary
Core of the parsing is reasonable; the blocker items are the Signal K path choice (attitude, and dropping invented accuracy paths), the timestamp source, and the broken `closeTo` assertions. Happy to help if anything is unclear.

msallin:
closed as stale (repo hygiene and overview reasons); happy when you reopen and discuss.

Issue #141, opened 2018-12-11
Remove moment from dependancies


joabakk:
Rewrite getTagBlock, GGA and GLL to avoid using moment, using code similar to `var s = new Date(1331209044000).toISOString();`

msallin:
`moment` is no longer a runtime dependency (see `package.json`). `ZDA.js`, `GGA.js`, `GLL.js`, and `lib/getTagBlock.js` now use `@signalk/nmea0183-utilities` `utils.timestamp` and native `Date`. The only remaining reference is `bench/moment.js` (performance comparison). Closing.

PR #281, opened 2026-04-11
Drop moment-timezone from sentence parse hot path

msallin:
## Summary

`moment-timezone` was called per-sentence in `GGA`, `GLL`, `GNS`, `RMC`, and in `lib/getTagBlock` to format or parse timestamps. Every call was replaceable with a native `Date` operation, and the library was imported but unused in `RMB` and `ZDA`. This PR drops the `moment-timezone` direct dependency and replaces each call site.

At typical NMEA rates (several Hz per sentence type, sustained) these calls sit squarely in the parse hot path. On a Raspberry Pi running a Signal K server on battery power, moment's overhead is disproportionate to what it actually does here (all it returns is today's UTC date as `DDMMYY`, or an ISO string from a unix timestamp — both trivially available via native `Date`).

## Changes

| File | Before | After |
|---|---|---|
| `hooks/GGA.js` | `utils.timestamp(time, moment.tz('UTC').format('DDMMYY'))` | `utils.timestamp(time)` |
| `hooks/GLL.js` | same | same |
| `hooks/GNS.js` | same | same |
| `hooks/RMC.js` | `moment.tz(timestamp, 'UTC').unix()` | `Math.floor(Date.parse(timestamp) / 1000)` |
| `hooks/RMB.js` | dead `require('moment-timezone')` | removed |
| `hooks/ZDA.js` | dead `require('moment-timezone')` | removed |
| `lib/getTagBlock.js` | `moment.tz(ms, 'UTC').toISOString()` | `new Date(ms).toISOString()` |
| `package.json` | `moment-timezone` direct dep | removed |

### Why `utils.timestamp(time)` with no date arg is equivalent

Looking at `@signalk/nmea0183-utilities`:

```js
exports.timestamp = function (time, date) {
  // ...
  if (date) {
    day = this.int(date.slice(0, 2), true)
    month = this.int(date.slice(2, 4), true)
    year = 2000 + this.int(date.slice(4, 6), true)
  } else {
    var dt = new Date()
    year = dt.getUTCFullYear()
    month = dt.getUTCMonth() + 1
    day = dt.getUTCDate()
  }
  var d = new Date(Date.UTC(year, month - 1, day, hours, minutes, seconds))
  return d.toISOString()
}
```

Passing `moment.tz('UTC').format('DDMMYY')` just formats today's UTC date as a 6-character string, which `utils.timestamp` then slices back into three integers. Omitting the argument entirely reaches the same three integers via `new Date().getUTC*()` directly. Identical output, one call path shorter, no moment allocation.

### Why `Math.floor(Date.parse(x) / 1000)` is equivalent to `moment.tz(x, 'UTC').unix()`

`Date.parse` returns milliseconds-since-epoch as a plain number without allocating a `Date` object. `.unix()` from moment returns integer seconds-since-epoch. Dividing by 1000 and flooring gives the same integer. No timezone data is involved because `Date.parse` of an ISO-8601 string with a `Z` suffix is already UTC.

## Benchmark

A micro-benchmark is included at `bench/moment.js`. It parses representative sentences through the existing `Parser` in a tight loop, measures over 5 trials with 5000-iteration warmup per sentence, and reports the min trial (least affected by GC/scheduler noise) and median trial.

Environment: Node v23.9.0, win32/x64, 200k iterations per trial.

| Sentence | Before (min ns/op) | After (min ns/op) | Before (median) | After (median) | Speedup (min) |
|---|---:|---:|---:|---:|---:|
| GGA | 6,924 | 3,940 | 7,232 | 4,419 | **1.76x** |
| GLL | 5,222 | 3,181 | 5,354 | 3,528 | **1.64x** |
| GNS | 5,928 | 4,223 | 6,357 | 4,497 | **1.40x** |
| **RMC** | **13,566** | **4,010** | **15,095** | **4,245** | **3.38x** |
| TAG | 4,513 | 2,655 | 4,661 | 3,060 | **1.70x** |

RMC shows the biggest speedup (~70% CPU reduction per sentence) because it had two moment calls per parse. The others had one moment call each.

To reproduce on your own hardware:

```bash
git checkout master
node bench/moment.js 200000 5    # before (needs moment-timezone installed)
git checkout perf/remove-moment-timezone
npm install                       # removes moment-timezone
node bench/moment.js 200000 5    # after
```

These numbers are from x86_64. On ARM (Raspberry Pi 3/4/5) absolute ns/op will be higher, but the speedup *ratios* should be similar since the moment/Date cost asymmetry is algorithmic, not architecture-dependent.

## Test plan

- [x] `npm test` passes: 238 passing, 0 failing (unchanged from master)
- [x] `test/RMC.js` asserts the exact ISO timestamp `'2014-04-03T08:54:12.000Z'` and exact unix seconds `1396515252` for the age field. Both pass, proving byte-for-byte compatibility of the RMC replacement.
- [x] `test/tagblock.js` asserts the exact ISO string `'2015-08-02T04:28:17.000Z'` from a unix epoch input. Passes, proving byte-for-byte compatibility of the `getTagBlock` replacement.
- [x] `npm ls moment-timezone` returns empty after the change.

## Notes

- `moment` (the core package, not `moment-timezone`) remains as a transitive dev/test dependency via `@signalk/signalk-schema` -> `tv4-formats`. It is not loaded by the runtime parse path on a Signal K server. The install tree shrinks by the moment-timezone data files (~900 KB).
- The `bench/` directory ships with the package since this repo has no `.npmignore` or `files` whitelist. If you would prefer the bench not to ship to npm, I am happy to add a `.npmignore` in a follow-up, but that felt out of scope for this change.
- I preserved an explanatory comment in `hooks/GGA.js` mentioning the previous `moment.tz('UTC').format('DDMMYY')` call, so a future reader understands why `utils.timestamp` is called without its second argument. Happy to remove it if you prefer.


msallin:
fixes #141

PR #304, opened 2026-04-17
test: push coverage to ~99.8% with mutation-testing-validated tests

msallin:
## Summary

Pushes test coverage to **99.74% statements / 99.31% branches / 99.74% lines** (from 96.14% / 82.62% / 96.14%). Test count: **266 → 423**.

Tests were validated with [StrykerJS](https://stryker-mutator.io/) mutation testing and iterated until the surviving mutations were either benign (debug strings, `.trim()` on already-clean data) or fundamentally unreachable. StrykerJS config and artifacts are not committed.

## Safety

**Every production-code change in this PR is strictly behavior-preserving** under the surface area exercised by the existing + new tests. I verified this by running the full test suite with every production file reverted to `upstream/master` while keeping the new tests from this PR:

| Config | Result |
|---|---|
| New tests + original production code (`upstream/master`) | 423/423 passing |
| New tests + refactored production code (this PR) | 423/423 passing |

Per-file rationale for each refactor is in [this comment](https://github.com/SignalK/nmea0183-signalk/pull/304#issuecomment-4269698802).

## What's covered

- Every hook file at 100% statement/branch coverage except `hooks/VDM.js`. The remaining gap (lines 461–472 `horvisib` emission, 497–498 all-filtered fallback) requires hand-crafted AIS type 8 DAC=1 FID=31 bit payloads and was left as-is.
- All `hooks/proprietary/*`, `hooks/seatalk/*`, and `lib/*` at 100% statements.

## Production-code refactors (all behavior-preserving, see per-file rationale comment)

- `hooks/seatalk/0x00.js`: remove dead mode/flag variables never read after assignment.
- `hooks/seatalk/0x84.js`: replace unreachable switch default with a ternary.
- `hooks/RMB.js`: drop `isNaN` guards on `utils.float()` results — `utils.float` never returns NaN.
- `hooks/proprietary/PBVE.js`: remove dead `highRpmAlarm` / `gaugeAlarmOn` in the B branch.
- `hooks/BWC.js`: remove always-true `values.length > 0 ? result : undefined` fallback.
- `hooks/proprietary/PNKEP.js`, `hooks/VTG.js`: remove unused `isEmpty` helpers.
- `hooks/GGA.js`, `GLL.js`, `GNS.js`, `DSC.js`, `VWR.js`, `ZDA.js`, `DBT.js`, `HDM.js`, `HDT.js`, `HDG.js`: simplify `isEmpty` / inline `typeof` guards — the `!== 'number'` branch was dead since `parts` are always strings.

## Test plan

- [x] `npm test` — 423 passing
- [x] `npm run coverage` — 99.74% stmts / 99.31% branches
- [x] `npm run prettier:check` — clean
- [x] Full test suite passes with production code reverted to `upstream/master` (verifies no behavior change).
- [x] StrykerJS mutation score ~80%; surviving mutants reviewed and classed as acceptable (debug strings, trim-on-clean-data, a few defensive branches).

### DIFF
--- a/hooks/BWC.js
+++ b/hooks/BWC.js
@@ -98,5 +98,5 @@ module.exports = function BWCHook(input) {
     })
   }
 
-  return values.length > 0 ? result : undefined
+  return result
 }
--- a/hooks/DBT.js
+++ b/hooks/DBT.js
@@ -72,6 +72,4 @@ module.exports = function (input) {
   return delta
 }
 
-const hasNoValue = (value) =>
-  (typeof value !== 'string' && typeof value !== 'number') ||
-  (typeof value === 'string' && value.trim() === '')
+const hasNoValue = (value) => typeof value !== 'string' || value.trim() === ''
--- a/hooks/DSC.js
+++ b/hooks/DSC.js
@@ -21,10 +21,7 @@ const debug = require('debug')('signalk-parser-nmea0183/DSC')
 const utils = require('@signalk/nmea0183-utilities')
 
 function isEmpty(mixed) {
-  return (
-    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
-    (typeof mixed === 'string' && mixed.trim() === '')
-  )
+  return typeof mixed !== 'string' || mixed.trim() === ''
 }
 
 function parsePosition(line) {
--- a/hooks/GGA.js
+++ b/hooks/GGA.js
@@ -59,10 +59,7 @@ Field Number:
 */
 
 function isEmpty(mixed) {
-  return (
-    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
-    (typeof mixed === 'string' && mixed.trim() === '')
-  )
+  return typeof mixed !== 'string' || mixed.trim() === ''
 }
 
 module.exports = function (input) {
--- a/hooks/GLL.js
+++ b/hooks/GLL.js
@@ -37,10 +37,7 @@ Field Number:
 */
 
 function isEmpty(mixed) {
-  return (
-    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
-    (typeof mixed === 'string' && mixed.trim() === '')
-  )
+  return typeof mixed !== 'string' || mixed.trim() === ''
 }
 
 module.exports = function (input) {
--- a/hooks/GNS.js
+++ b/hooks/GNS.js
@@ -65,10 +65,7 @@ Field Number:
 */
 
 function isEmpty(mixed) {
-  return (
-    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
-    (typeof mixed === 'string' && mixed.trim() === '')
-  )
+  return typeof mixed !== 'string' || mixed.trim() === ''
 }
 
 const MODES = {
--- a/hooks/HDG.js
+++ b/hooks/HDG.js
@@ -32,10 +32,10 @@ Field Number:
 */
 
 function isEmpty(mixed) {
-  return (
-    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
-    (typeof mixed === 'string' && mixed.trim() === '')
-  )
+  if (typeof mixed === 'number') {
+    return false
+  }
+  return typeof mixed !== 'string' || mixed.trim() === ''
 }
 
 module.exports = function (input) {
--- a/hooks/HDM.js
+++ b/hooks/HDM.js
@@ -35,10 +35,7 @@ const utils = require('@signalk/nmea0183-utilities')
 module.exports = function (input) {
   const { id, sentence, parts, tags } = input
 
-  if (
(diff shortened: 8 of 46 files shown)


msallin:
## Per-file rationale for production-code changes

All production code changes in this PR are **strictly behavior-preserving refactorings** under the surface area exercised by the existing + new tests. I verified this by running the full test suite with every production file reverted to `upstream/master` while keeping the new tests: **423/423 tests pass in both configurations.**

Each change below explains what was removed or simplified and why it is safe.

### `hooks/seatalk/0x00.js` — drop dead mode/flag variables
The original function declared `mode`, `modeY`, `modeZ`, and a sign variable `s`, conditionally assigned them based on Y/Z flag bits, and then **never read any of them**. Only `depthbelowtransducer` is emitted. The removed block was stub scaffolding for alarm flags that were never wired into the output; the hook's observable output is unchanged.

### `hooks/seatalk/0x84.js` — switch default removed
`modeVar = Z & 0x2` can only produce `0` or `2` (bitwise AND with a single bit). The `default: break` case in the switch was unreachable. Replaced the switch with `mode = (Z & 0x2) === 0 ? 'standby' : 'auto'`, which produces the same result for every reachable input.

### `hooks/RMB.js` — drop `isNaN` guards on `utils.float` results
`@signalk/nmea0183-utilities`' `float()` is implemented as:
```js
exports.float = function (n) {
  const parsed = parseFloat(n)
  return Number.isNaN(parsed) ? 0.0 : parsed
}
```
It never returns `NaN`. So `bearing = !isNaN(bearing) ? bearing : 0.0` (and the distance/crossTrackError equivalents) is always a no-op. Removed those and the unused `-1` initial values for `latitude`/`longitude`. The VMG guard was compressed to `rawVmg > 0 ? rawVmg : 0.0`, preserving the clamp-negative-to-zero semantics.

### `hooks/proprietary/PBVE.js` — remove dead `gaugeAlarmOn` / `highRpmAlarm` in B branch
In the `productCode === 'B'` branch, `highRpmAlarm` is computed solely to feed `gaugeAlarmOn = highRpmAlarm > rpm ? 1 : 0`, and `gaugeAlarmOn` is never read — the emitted `values` for RH30/RH60/RH110 only contain `rpm` and `runTime`. This is stub scaffolding for an alarm flag that was never plumbed through (the D/E branches do emit it via `convertedValue.meta.gaugeAlarmOn`, but B does not). Emitted output unchanged.

### `hooks/BWC.js` — drop always-true `values.length > 0` fallback
`return values.length > 0 ? result : undefined`: the code pushes a `navigation.courseGreatCircle.nextPoint.position` entry in **both** branches of the lat/lon check (the real value in the `if`, `null` in the `else`), so `values.length` is always ≥ 1 and the `: undefined` branch is unreachable. Simplified to `return result`.

### `hooks/proprietary/PNKEP.js`, `hooks/VTG.js` — remove unused `isEmpty`
Both files defined an `isEmpty` helper but never called it. Pure dead code.

### `hooks/GGA.js`, `GLL.js`, `GNS.js`, `DSC.js`, `VWR.js`, `ZDA.js` — simplify `isEmpty`
Original:
```js
function isEmpty(mixed) {
  return (
    (typeof mixed !== 'string' && typeof mixed !== 'number') ||
    (typeof mixed === 'string' && mixed.trim() === '')
  )
}
```
Simplified to:
```js
function isEmpty(mixed) {
  return typeof mixed !== 'string' || mixed.trim() === ''
}
```
The behavior differs **only** when called with a number: the old form returned `false`, the new form returns `true`. But `parts` always come from `sentence.split(',')` in `lib/index.js`, so every value passed to these helpers is a string. The number branch was dead.

### `hooks/HDG.js` — keep number handling in `isEmpty`
HDG calls `isEmpty(deviationDir)` and `isEmpty(variationDir)` where those values are numbers (`1` or `-1`) set by the `=== 'E' ? 1 : -1` ternary. So the simplified form above would have changed behavior. I preserved the number path explicitly:
```js
function isEmpty(mixed) {
  if (typeof mixed === 'number') {
    return false
  }
  return typeof mixed !== 'string' || mixed.trim() === ''
}
```
Same output as the original for every reachable input.

### `hooks/HDM.js`, `HDT.js`, `DBT.js` — inline simplified typeof guard
The inline duplicate of the old `isEmpty` pattern. Simplified to `typeof x !== 'string' || x.trim() === ''` for the same reason as above: the `!== 'number'` branch was unreachable.

---

## Omitted from this PR (flagged for follow-up)

I had also included a small null-safety fix in `lib/index.js` — the constructor normalises `opts` into `this.options` on line 33 but then references raw `opts` on line 45, so `new Parser(null)` throws. I reverted this from the PR because it is a genuine behavior change (crash → no crash), not a behavior-preserving refactor, and it is out of scope for a test-coverage PR. Happy to open a separate PR for it if you want — it is a one-line change with one test.

PR #309, opened 2026-04-19
refactor: migrate to TypeScript strict

msallin:
Single-commit TypeScript-strict migration of the parser, sitting cleanly on top of [#308 (bump utilities to ^1.0.0)](https://github.com/SignalK/nmea0183-signalk/pull/308) which landed as `89ae374`.

No public-API change. Consumers of `@signalk/nmea0183-signalk` see the same parser with the same delta output.

## What moves where

- `index.js` / `lib/` / `hooks/` / `bin/` / `bench/` / `custom-sentence-plugin/` → `src/`, authored as strict TypeScript.
- `tsconfig.json` turns on `strict` + `noUncheckedIndexedAccess` + `noImplicitOverride` + `noImplicitReturns` + `noUnusedLocals` + `noUnusedParameters`.
- `tsconfig.build.json` compiles to `dist/` with `.d.ts`, `.d.ts.map`, `.js.map`.
- `package.json` `main` / `types` point at `dist/`. `prepublishOnly` runs a fresh build. `bin/nmea0183-signalk` remains a tiny JS shim re-requiring the compiled entry so the published CLI keeps working.
- Mocha runs `.ts` specs directly via `tsx/cjs` (`.mocharc.cjs`).
- Stryker config (added upstream in `85166eb`) now targets `src/**/*.ts` / `test/**/*.ts` and ignores `dist`.
- `src/types/` carries ambient declarations for the only two remaining untyped runtime deps — `ggencoder`, `split` — plus `signalk-signalk-schema`.

## Type-system adjustments for the v1 utilities

The utilities v1 tightened `Pole` to `'N' | 'S' | 'E' | 'W'` and `transform`'s units to `UnitFormat`. Raw `parts[N]!` strings from NMEA fields can't be narrowed to those unions without runtime validation, so 13 `utils.coordinate` call sites across `BWC` / `BWR` / `GGA` / `GLL` / `GNS` / `RMB` / `RMC` cast to `Pole` at the field boundary. Runtime behaviour for all valid sentences is unchanged; a malformed sentence (e.g. lowercase `'s'`) now throws `unsupported pole: s` from the utilities instead of silently returning a wrong-sign coordinate — preferable failure mode. `MWV`'s wind-unit selector is typed as `UnitFormat`.

## Verification

- [x] `npm ci && npm test` — 424 passing
- [x] `npm run typecheck` — clean under `strict` + `noUncheckedIndexedAccess`
- [x] `npm run build` — emits `dist/` with `.js`, `.d.ts`, `.d.ts.map`, `.js.map`
- [x] `npm run prettier:check` — clean
- [x] `npm run coverage` — within 1% of pre-migration baseline (source-map attribution noise from c8 reading through tsx, not a real regression)

## Test plan for reviewers

- [ ] `npm ci && npm test` on the branch
- [ ] `npm run typecheck` passes
- [ ] `npm run build` emits a clean `dist/`
- [ ] Spot-check a couple of hooks (e.g. `src/hooks/RMC.ts`, `src/hooks/APB.ts`, `src/hooks/seatalk/0x50.ts`) — conversion is mechanical, logic unchanged
- [ ] `npm pack --dry-run` produces an expected tarball (`dist/`, `bin/`, `README.md`, `LICENSE`)


msallin:
External-reviewer pass complete; pushed the fixes.

**Applied** (ship-stopper + worth-fixing):
- CI now runs `npm run typecheck` and `npm run build` on both Node 22 and 24, before the test/coverage steps. Without this a TS regression or broken `.d.ts` emission only surfaced at `prepublishOnly`.
- `package.json#exports` map with `types`/`require`/`default` conditions so `node16`/bundler resolvers work. `main`/`types` kept as legacy fallbacks.
- `engines.node: ">=22"` added (matches `signalk-server`).
- `sideEffects: false` (tree-shaking hint).
- Removed the stale `standard` config key (we use prettier + tsc now).
- Excluded `src/bench` and `src/custom-sentence-plugin` from `tsconfig.build.json` so neither lands in `dist/` or the tarball. Both were example/benchmark code that had no business shipping to consumers.
- Simplified `tsconfig.build.json` (dropped duplicate `outDir`/`declaration`/`declarationMap`/`sourceMap` already inherited from base).
- Removed the stale `deploy` / `release` / `create-release` npm scripts and the `@signalk/github-create-release` devDep — obsolete now that `publish.yml` handles OIDC-trusted publishing on tag push. One source of truth for release flow.
- Ran a CLI smoke-test against the compiled `dist/bin/nmea0183-signalk.js` to confirm shebang + exec bits work end-to-end.

**Deferred** (discussion / separate PR):
- The 71-file `module.exports = X; module.exports.default = X; module.exports.Parser = X` pattern propping up `esModuleInterop` compatibility. Worth cleaning up, but the only real consumer-visible risk is `signalk-server`'s `require()` pattern — I'd like to check that before ripping the scaffolding out.
- Moving the 13 `utils.coordinate(parts[N]!, parts[M]! as Pole)` casts into a `toPole()` / `toUnit()` boundary helper that turns malformed NMEA fields into skipped deltas instead of throwing up through the Parser. Material behaviour improvement, but properly separate PR.
- `test/chai-augment.d.ts` location (works today because it's in the TS include; would silently break under `isolatedModules` or narrower globs — follow-up to move into a `test/setup.ts` that mocha loads via `--require`).
- 329 `as any` casts on `parse()` return in tests — ergonomic improvement via a tiny typed helper, not behaviour.
- `src/lib/index.ts` imports `../../package.json` for `name`/`version`/`author`/`license`; with `rootDir: src` tsc tolerates it but the path leaves a whiff. Extract to a typed `pkg-info.ts` or read via `require` at runtime; non-blocking.
- `src/hooks/VDM.ts` has a handful of `var` declarations (pre-existing).

Final state: 424 tests pass, `tsc --noEmit` clean, `npm run build` emits `dist/` clean, prettier clean, tarball 505.4 kB / 295 files. CLI smoke-test parses an `$GPRMC` sentence end-to-end and emits a valid Signal K delta.

msallin:
Applied the larger deferred items plus downstream-compat verification. Pushed `232fd07`.

## Applied

**Cleaned up the 71-file dual-export pattern.** Every hook, lib file, and bin file used to end with:
```ts
export default X
module.exports = X
module.exports.default = X
// (+ module.exports.Parser = Parser in src/lib/index.ts)
```
Those three manual lines propped up `esModuleInterop` + raw-`require()` back-compat. Internal modules (hooks, lib helpers) don't need it — they're only consumed by `import X from './X'` which tsc compiles correctly via `exports.default`. Stripped the manual lines from 68 files. **`src/index.ts` kept** the full dual-export because it's the public entry and signalk-server / legacy consumers may do `const Parser = require('@signalk/nmea0183-signalk')` expecting the constructor directly.

**Added `src/lib/nmea-casts.ts`** with `toPole()` / `toUnit()` / `coord()` / `magVar()` boundary helpers. Replaced the 13 `utils.coordinate(parts[N]!, parts[M]! as Pole)` call sites across BWC / BWR / GGA / GLL / GNS / RMB / RMC with `coord(parts[N]!, parts[M]!)` which returns `null` for unknown pole letters rather than throwing. Material behaviour improvement: a malformed sentence with lowercase `'s'` now emits `value: null` in the delta (SignalK convention for "unknown") instead of bubbling a `TypeError` up through the Parser. Same treatment for the one magnetic-variation call site in RMC (`magVar()` wrapper).

**Removed `var` declarations in `src/hooks/VDM.ts`** (5 sites → `let` / `const` depending on mutation).

## Deferred (low-value, not blocking)

- Moving `test/chai-augment.d.ts` to a `--require`-loaded setup file. Current wiring works via tsconfig `include`; reviewer noted it would silently break under future tsconfig narrowing, but that's hypothetical.
- Replacing ~329 `as any` on `parse()` return in tests with a typed helper. Mechanical but large diff; better as a dedicated test-ergonomics PR.
- Refactoring the `import pkg from '../../package.json'` in `src/lib/index.ts`. The `rootDir: src` mix works today; tsc doesn't emit package.json to dist. Value is marginal.

## Downstream verification (signalk-server)

Ran the fork against two signalk-server versions via Docker with the repo mounted over `node_modules/@signalk/nmea0183-signalk`:

- **`cr.signalk.io/signalk/signalk-server:latest`** — server starts clean (no errors on import), and a direct `new Parser().parse(...)` test from inside the signalk-server install produces correct deltas on RMC, GGA, and AIVDM sentences.
- **`cr.signalk.io/signalk/signalk-server:v2.20.0`** (roughly the line shipping on Victron Cerbo GX at time of writing) — same result: server boots clean, parser works identically, 0 errors in logs.

Both server versions accept the dual-export shape (`typeof Parser === 'function'`, `Parser.default` and `Parser.Parser` both point at the constructor), so every documented consumer pattern works: `const P = require('@signalk/nmea0183-signalk')`, `import P from '...'` (with `esModuleInterop`), and `require('...').Parser`.

## Final state

424 tests pass, typecheck clean, build clean, prettier clean, tarball 505.4 kB / 295 files, CLI smoke-test green, integration verified on SK latest + v2.20.0.

PR #334, opened 2026-06-16
fix(VWR): fall back to m/s and km/h speed fields

johansolve:
## Problem

The VWR hook only reads the **knots** field (`parts[2]`) for `environment.wind.speedApparent`, ignoring the **m/s** (`parts[4]`) and **km/h** (`parts[6]`) fields.

The VWR sentence may report wind speed in any of the three units:

```
        0     1   2   3   4    5   6     7  8
$--VWR,x.x,  a,  x.x, N, x.x,  M, x.x,   K*hh
              L/R knots    m/s     km/h
```

Some talkers leave the knots field empty and populate only m/s. A real example from a Brookhouse multiplexer:

```
$IIVWR,154.0,R,,,7.0,M,,*4B
```

Here the actual apparent wind is 7.0 m/s, but since `utils.float('')` returns `0`, the parser reported `environment.wind.speedApparent: 0`. With the device also emitting MWV (which parses correctly) under the same `$source`, the value flickered between the correct speed and 0.

Note this is the mirror of the case already handled in #122, which made VWR accept sentences where fields 4–7 (m/s/km/h) are empty — i.e. knots-only. The knots-empty / m/s-present case was never covered.

## Fix

Fall back **knots → m/s → km/h** for `speedApparent`. When all three speed fields are empty, omit `speedApparent` from the delta entirely rather than reporting a misleading `0`. `angleApparent` is unchanged.

### Behaviour change

A VWR sentence that passes the `empty > 4` guard but has all three speed fields empty (e.g. `$IIVWR,154.0,R,,N,,M,,K`) previously emitted `speedApparent: 0`; it now omits the path. This is the intended correctness improvement (no data vs. a wrong `0`).

## Tests

Added cases for m/s-only, km/h-only, and the all-empty-speed-fields omission. Full suite passes (427 tests), typecheck and prettier clean.

### DIFF
--- a/src/hooks/VWR.ts
+++ b/src/hooks/VWR.ts
@@ -57,25 +57,42 @@ const VWR: HookFn = function (
     rightPositive = -1
   }
 
+  // The speed is reported in up to three units; some talkers leave the knots
+  // field (parts[2]) empty and only populate m/s (parts[4]) or km/h (parts[6]).
+  // Fall back to those instead of reporting 0.
+  let speedApparent: number | undefined
+  if (!isEmpty(parts[2])) {
+    speedApparent = utils.transform(utils.float(parts[2]!), 'knots', 'ms')
+  } else if (!isEmpty(parts[4])) {
+    speedApparent = utils.float(parts[4]!)
+  } else if (!isEmpty(parts[6])) {
+    speedApparent = utils.float(parts[6]!) / 3.6
+  }
+
+  const values = [
+    {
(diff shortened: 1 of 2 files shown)


johansolve:
Please label as fix

msallin:
@johansolve please squash the 3 commits into one and rebase on latest master. Then I'm good to go.

johansolve:
@msallin  Squashed and rebased on latest master 👍

Issue #217, opened 2022-01-03
DSC Distress Alerts


TwoCanPlugIn:
Is SignalK meant to parse NMEA 183 DSC and DSE sentences ?

Sending the following sentences does not generate any SIgnalK nodes as far as I can tell:

$CDDSC,12,5031105200,,05,00,2380814428,1800,,,R,E*6C
$CDDSE,1,1,A,5031105200,04,19252411*18

and the following error is logged.

Jan 03 21:55:07 signalk:streams:nmea0183-signalk [error] line is not defined +4h

TwoCanPlugIn:
Some further observations. 

BTW, Not sure what happened with the formatting of the NMEA 183 sentences above. Have now learnt to enclose as code.

The DSC Distress alert sentences were 
`$CDDSC,12,5031105200,,05,00,2380814428,1800,,,R,E*6C`
`$CDDSE,1,1,A,5031105200,04,19252411*18` 

When sent over TCP or UDP no other nodes/contexts (or whatever you call them) are generated.

Similarly for a DSC Position Report, using the following sentence, nothing is generated.
`$CDDSC,20,5031105200,08,21,26,2380814428,1800,,,B,E*77`
`$CDDSE,1,1,A,5031105200,04,19252411*18`

Instead if a DSC Distress Alert Relay is sent, using the following sentence:
`$CDDSC,16,2350763930,12,12,00,2380814428,1800,5031105200,05,S,E*65`
`$CDDSE,1,1,A,5031105200,04,19252411*18`

the following context is created with the following values:

`vessels.urn:mrn:imo:mmsi:235076393

navigation.position
{
  "latitude": -38.13333333333333,
  "longitude": 144.46666666666667
}

notifications.epirb
{"message":"DSC Distress Recieved! Nature of distress: epirb"}`

The values in this Distress Relay sentence should be interpreted as:
Format Specifier 16 - All Ships
MMSI Address - 235076393 (This is the MMSI Address of the relay vessel)
DSC Category 12 - Distress
First Telecommand 12 - Distress Relay
Type of Communications 00 - F3E/G3E, All modes TP 
Vessel In Distress 503110520 (this is the actual vessel in distress)
Nature of Distress 05 - Sinking

Obviously this is being parsed incorrectly, using the wrong MMSI address for the vessel in distress, and the wrong value for nature of distress (12 being interpreted as MOB).

This is just totally broken. 

No need to describe its impact on OpenCPN.

[...]

PR #336, opened 2026-06-17
Recognize DSC distress alerts with an implied (empty) category

clarkbw:
A distress alert uses format specifier 112 and, per ITU-R M.493, carries no category field — it's implied. #315 stopped these alerts from being dropped by the entry guard, but the category `switch` still doesn't recognize them, so a spec-correct distress alert surfaces only as "DSC Message Not Handled" instead of its distress notification.

This falls back to the format specifier when the category field is empty, completing the FS-driven distress dispatch that #315 explicitly deferred. It also guards position emission on a full 10-digit position field — a distress alert without a position previously yielded `NaN` latitude/longitude.

Rebased on master after #315; the sparse-field and missing-MMSI hunks this PR originally carried are now covered upstream. Refs #217.

### DIFF
--- a/src/hooks/DSC.ts
+++ b/src/hooks/DSC.ts
@@ -85,6 +85,14 @@ const DSC: HookFn = function (
     return null
   }
 
+  // A DSC call with no sender MMSI cannot be attributed to a vessel, so there
+  // is no context to publish under. Drop it rather than emitting a delta with
+  // an empty `vessels.urn:mrn:imo:mmsi:` context — this also guards the MMSI
+  // substring below against a missing field (a bare `$CDDSC,12` would throw).
+  if (isEmpty(parts[1])) {
+    return null
+  }
+
   // for some reason, it seems the sender identification is mmsi+'0', so we
   // strip the trailing zero to get a 9 digit mmsi
   var mmsi = parts[1]!.substring(0, 9)
--- a/src/hooks/DSC.ts
+++ b/src/hooks/DSC.ts
@@ -66,6 +66,12 @@ const DSC: HookFn = function (
   const { sentence, parts, tags } = input
   var values: Array<{ path: string; value: unknown }> = []
 
+  // A distress alert uses format specifier 112 and, per ITU-R M.493, carries
+  // no DSC category field — it is implied. Recognize distress from the format
+  // so these alerts are neither dropped by the sparse-field guard nor
+  // mis-classified as unhandled (#217).
+  const isDistressFormat = parts[0] === '12'
+
   const empty = parts.reduce((e, val) => {
     if (isEmpty(val)) {
       ++e
@@ -73,7 +79,9 @@ const DSC: HookFn = function (
     return e
   }, 0)
 
-  if (empty > 3) {
+  // A received distress alert must never be silently discarded, even when most
+  // of its fields are blank.
+  if (empty > 3 && !isDistressFormat) {
     return null
   }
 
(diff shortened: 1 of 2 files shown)


src/hooks/DSC.ts
```diff
@@ -66,14 +66,22 @@ const DSC: HookFn = function (
...
     if (isEmpty(val)) {
       ++e
     }
     return e
   }, 0)
 
-  if (empty > 3) {
+  // A received distress alert must never be silently discarded, even when most
+  // of its fields are blank.
+  if (empty > 3 && !isDistressFormat) {
```
msallin: Exempting distress from this guard makes the unguarded `parts[1]!.substring(0, 9)` below reachable for blank frames. `$CDDSC,12,,,07,,,,,,*57` (empty MMSI) now emits a delta with an empty `vessels.urn:mrn:imo:mmsi:` context instead of `null`. Suggest `if (isEmpty(parts[1])) return null` before the MMSI line; it also fixes the pre-existing TypeError on a bare `$CDDSC,12`, and breaks no existing test.

src/hooks/DSC.ts
```diff
@@ -158,7 +170,7 @@ const DSC: HookFn = function (
     }
   })*/
 
-  if (get_position) {
+  if (get_position && /^\d{10}$/.test(parts[5] ?? '')) {
```
msallin, line 177: Good NaN guard. The regex rejects both empty and non-10-digit values, but only the empty `parts[5]` case is tested. Worth adding a distress case with a non-empty, non-10-digit position (e.g. a truncated field) asserting no `navigation.position` is emitted, so a later loosening of this regex can't silently reintroduce NaN coordinates.

msallin (commented):
Two inline notes on the major points from a review pass.

src/hooks/DSC.ts
clarkbw: Confirmed `$CDDSC,12,,,07,,,,,,*57` was emitting an empty context. Added `if (isEmpty(parts[1])) return null` before the MMSI substring — also kills the bare-`$CDDSC,12` TypeError. While there I noticed a sparse frame with a *non-numeric* MMSI (`$CDDSC,12,ABC,...`) still slips through to a malformed context, since it's not empty — happy to tighten `parts[1]` to a numeric check in this PR, or leave it as a follow-up?

src/hooks/DSC.ts
clarkbw: Good call — I added a truncated case first but realized it doesn't actually pin the anchors (a too-short field is rejected either way, so it never detects a loosening). Replaced it with a small table — too-short, too-long (11 digits), and a non-digit prefix — so dropping `^` or `$` now fails a test. Also added a test that the implied-category fallback only fires for format 112, not any blank-category frame.

src/hooks/DSC.ts
clarkbw: Opened #343 for the non-numeric MMSI (`$CDDSC,12,ABC,...`) case — it's a malformed frame rather than the sparse-but-valid case this PR is about, so I kept it out of scope. The `isEmpty` guard already covers the empty context + bare-`$CDDSC,12` TypeError you flagged.

Anything else you'd want before this can land? Both your review notes are addressed.

clarkbw:
Rebased on master after #315 landed — it covered the sparse-field and missing-MMSI hunks this PR originally carried, so what's left is just the implied-category dispatch #315 deferred. I updated the existing empty-Category test from asserting `notifications.dsc_parser` to `notifications.sinking` accordingly.

msallin:

PR #337, opened 2026-06-17
Refine DSC position from the $--DSE expansion sentence

clarkbw:
A `$--DSE` sentence refines the position of a preceding `$--DSC` from whole minutes to ten-thousandths of a minute, but it carries only the fractional component — never absolute coordinates. There was no DSE hook, so this finer position was discarded entirely.

This remembers each DSC position in the parser session keyed by MMSI, then adds a DSE hook that recalls it and emits the refined `navigation.position` under the same context. DSE sentences without a preceding DSC fix, and multi-sentence groups, are ignored. Tests included.

Touches `DSC.ts` to store the position in the session, so it overlaps slightly with #336 — whichever merges first, the other rebases cleanly.

### DIFF
--- a/src/hooks/DSC.ts
+++ b/src/hooks/DSC.ts
@@ -16,6 +16,7 @@
  */
 
 import type { Delta, HookFn, ParserInput, ParserSession } from '../types'
+import { rememberDscPosition } from './dscSession'
 import Debug from 'debug'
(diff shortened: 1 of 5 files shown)


clarkbw:
updated, this is ready for review whenever

msallin:

PR #344, opened 2026-07-11
Attribute DSC distress relays to the casualty and read the real nature of distress

clarkbw:
While testing DSC handling against a radio that forwards mayday relays, I found the hook mis-parses distress *relays* — an all-ships (116), individual (120) or area (102) call carrying the distress category. Worked example (a coast station relaying an EPIRB activation):

```
$CDDSC,16,0031600010,12,112,00,1423108312,2019,3162009110,12,,*47
```

Three things go wrong today:

1. **Nature of distress is read from the wrong field.** In a relay, field 3 is the relay telecommand (`112`), not a nature code — so every relay resolves to `notifications.unassigned`. The real nature is in field 8 (`12` = EPIRB here).
2. **The distress is attributed to the wrong vessel.** The position in field 5 and the distress itself belong to the *casualty* (field 7, `316200911`), but the delta is published under the relaying station's context (`003160001`). A consumer plotting the distress would mark the coast station's MMSI at the casualty's position.
3. **The acknowledgement char (field 9, R/B/S) is dropped.**

This PR teaches the distress branch to detect a relay (category 12 with a non-112 format specifier): nature comes from field 8 via the extracted `natureOfDistress()` (the same switch, now shared with the first-party path), the delta context switches to the casualty (trailing zero stripped, same as field 1), the notification message names the relaying station, and the ack char is appended to the message when present. If the casualty MMSI is absent (vessel unknown — legitimate for a relay), it falls back to the relaying station's context rather than dropping a distress. First-party alerts (format 112) are untouched — all existing tests pass unmodified.

Heads up: this overlaps #336 in the same `switch` — happy to rebase whichever lands second.

https://claude.ai/code/session_01GpBXJ9buMsou2Fxt5heT1x

### DIFF
--- a/src/hooks/DSC.ts
+++ b/src/hooks/DSC.ts
@@ -59,6 +59,38 @@ function parsePosition(line: string): { longitude: number; latitude: number } {
   return { longitude: lon_dec, latitude: lat_dec }
 }
 
+function natureOfDistress(code: string | undefined): string {
+  switch (code) {
+    case '00': // = Fire, explosion
+      return 'fire'
+    case '01': // = Flooding
+      return 'flooding'
+    case '02': // = Collision
+      return 'collision'
+    case '03': // = Grounding
+      return 'grounding'
+    case '04': // = Listing, in danger of capsize
+      return 'listing'
+    case '05': // = Sinking
+      return 'sinking'
+    case '06': // = Disabled and adrift
+      return 'adrift'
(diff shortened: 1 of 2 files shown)


msallin:
Merged into master as a84abd3 (squash). The merge call errored on GitHub's side, so this PR was never marked as merged; closing it manually. Thanks @clarkbw!

msallin:

PR #368, opened 2026-09-13
fix(DSC): validate MMSI fields and label distress acknowledgements

### LINKED ISSUE #343: DSC: sparse frame with a non-numeric sender MMSI slips through to a malformed context
Follow-up from #336 (review thread [r3445099496](https://github.com/SignalK/nmea0183-signalk/pull/336#discussion_r3445099496)).

#336 added `if (isEmpty(parts[1])) return null` before the MMSI substring, which drops empty-MMSI frames (`$CDDSC,12,,,07,,,,,,*57`) and the bare-`$CDDSC,12` TypeError. But a sparse frame with a *non-empty but non-numeric* MMSI still gets through:

```
$CDDSC,12,ABC,...
```

`isEmpty` is false, so `parts[1]!.substring(0, 9)` runs on `ABC` and produces a malformed `vessels.urn:mrn:imo:mmsi:ABC` context.

Fix is to tighten the `parts[1]` guard from empty-check to a numeric check (MMSI is always digits per ITU-R M.493). Left out of #336 to keep that PR scoped to distress-alert recognition.

msallin:
Follow-up to #336, #344 and #337.

- **MMSI validation**: a sender MMSI must be 9 or 10 digits, otherwise the sentence is dropped. A non-numeric value produced a malformed `vessels.urn:mrn:imo:mmsi:ABC` context. A non-numeric casualty MMSI in a relay now falls back to the relaying station, like an absent one. Closes #343.
- **Acknowledgements**: a distress acknowledgement (telecommand 110) was reported as a relay. It is now reported as an acknowledgement, still attributed to the casualty.
- **Polish**: the distress notification message read "Recieved" and now reads "Received"; anything matching on the old text needs updating. The DSC field documentation now covers fields 7 to 9, and the README lists DSE.

Each change is covered by tests; the full suite, typecheck, prettier and build pass.


msallin:
Merged into master as 1199140 (squash). The merge call errored on GitHub's side, so this PR was never marked as merged; closing it manually.

msallin:

PR #373, opened 2026-09-18
fix: correct ten miscalculated and dropped hook conversions

KEGustafsson:
* fix: correct seven miscalculated hook conversions

Each of these was verified against the primary specification and covered by a regression test.

SeaTalk 0x20 added the two speed bytes instead of combining them into a 16-bit value. The two readings agree below 25.6 knots, where the high byte is always zero, which is why it went unnoticed; above that the speed was badly under-reported. The byte order follows 0x26 and 0x52, whose little-endian reading is exercised in the field.

SeaTalk 0x84 and 0x9C computed the compass heading's last term from bit 0 of U. Knauf's reference defines it in prose as the number of bits set in the two higher bits of U, and the C-like shorthand printed beside that prose does not implement its own prose: `==` binds tighter than `&`, so `U & 0xC == 0xC` collapses to `U & 1`. Transliterating the shorthand put a one degree error on six of the sixteen values U can take. The formula now lives in one place shared by both datagrams.

SeaTalk 0x84 and 0x9C also dropped any field that was exactly zero, because the values were tested for truthiness. A rudder amidships and a heading of due north were never reported, leaving stale values on displays.

SeaTalk 0x85 assembled the cross track error with the low nibble at the top. Knauf's worked example is 2.61 nm, which encodes as 0x105 over X6 XX = 5_ 10, and that was being reported as 12.96 nm. The range field alongside it already read the nibbles in the right order.

AIS meteorological messages report salinity in per mille, so seawater reaches the conversion table as about 35. It was divided by 100 as if it were a percentage, overstating the ratio tenfold.

MWV left the wind speed unit letter S out of its table, so statute miles per hour passed through unconverted as metres per second.

ZDA truncated its 4-digit year to two digits and handed it to a helper that re-expands 2-digit years with the IEC 61162-1 pivot, moving anything before 1980 forward a century: a 1979 sentence was stamped 2079. The year is now read as sent, with the pivot kept for talkers that send only two digits.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016BfAzTXR83R1XynMDKxGLv

* fix: honour RMC status, unit fallbacks, and constellation satellite ids

The three findings left over from the hook review, on files the earlier branch does not touch.

RMC ignored its status field. Field 1 is the receiver status, 'A' active or 'V' void, and a void sentence still carries whatever the receiver last held in its registers. Position, course, speed and variation were published from those registers as though they were a live fix. They are now reported as unavailable instead, which is how GLL already treats its own void status. The datetime is kept, since a receiver that has lost its fix normally still keeps good time and RMC is a common clock source.

DBT, DBK and DBS each carry the sounding three times over, in feet, metres and fathoms, and a talker is free to populate only the unit its operator selected. DBK and DBS read the metres field alone, so a sounder configured for feet or fathoms produced no depth at all; DBT covered feet but not fathoms. All three now share one helper that prefers metres and falls back through feet and fathoms, using exact conversion factors rather than the truncated ratios in the utilities package.

VHW carries the speed in knots and in km/h. Only the knots field was read, so a talker populating only km/h reported no speed.

GSV subtracted 64 from every satellite identifier of 64 or more. That offset belongs to GLONASS alone, which NMEA numbers 65-96 as slot plus 64; applying it everywhere renumbered QZSS satellites, which run 193-202, onto GPS identifiers, and did the same to SBAS and to BeiDou above 400. The offset is now removed only for talkers that can carry GLONASS and only within its range.


tkurki:
The changes that are not related should be separate commits.

KEGustafsson:
I'll splitt per hook. Closing this one now.

